Issue #17, opened 2020-07-20
Building docker image with goreleaser to inject variables from ldflags


rockspore:
The docker image building process at this point does not provide those variables since it does not use goreleaser.

PR #20, opened 2020-07-24
Building docker image with goreleaser to inject variables from ldflags #17

theganyo:


.github/workflows/docker.yaml
```diff
@@ -21,60 +25,66 @@ on:
...
     runs-on: ubuntu-latest
     steps:
+      - run: echo "::set-env name=VERSION::$(echo ${GITHUB_REF//\//-} | cut -d'-' -f3-)"
+      - run: echo "::set-env name=DATE::$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
       - 
         name: Check out code
         uses: actions/checkout@v1
       -
-        name: Build and Push Docker image
+        name: Ubutu image
```
rockspore, line 60: Ubuntu ;-)

rockspore:
The local release is commented out in `release.yaml` right now. Should we uncomment that to resume building the binary?

theganyo:
No, we don't want to enable the local release... we only want it to run on goboring in Github. I left it there as reference in case we need to run it locally for some reason.

PR #41, opened 2020-08-13
Added tests scripts to be used locally as well as workflows in integration test later on.

rockspore:
No need to merge now. Just to open it here for easier diff.


rockspore:
Working on a `README.md` for the tests.

PR #54, opened 2020-09-01
Supporting analytics uploading to UAP directly with given service account credentials

### LINKED ISSUE #42: NG SaaS analytics upload
In NG SaaS the Apigee runtime instance and the Apigee Adapter inevitably reside in different clusters. So the adapter has to find a different way of sending analytics now that the UDCA endpoint in the runtime is unreachable.

Either a UDCA is deployed along with the adapter or a "legacy" way is used. More investigation is on the way.

rockspore:
Fix #42 

A service account with role Apigee Analytics Agent can be specified during provision. The credentials will be encoded into a k8s Secret in the generated config.yaml.

Upon starting, the adapter attempts to read the credentials from the config.yaml and then falls back to the secret path (`--analytics-secret` default to `/analytics-secret`). If credentials end up being unavailable, it will validate either the internalAPI or fluentd endpoint for analytics uploading.


server/config.go
```diff
@@ -170,12 +179,18 @@ func (c *Config) Load(configFile, policySecretPath string) error {
 			kidProps, _ = base64.StdEncoding.DecodeString(secret.Data[SecretPropsKey])
 			jwksBytes, _ = base64.StdEncoding.DecodeString(secret.Data[SecretJKWSKey])
 
+			// TODO: DecodeString() never returns nil even on error
+			// the following check is not effective
```
theganyo, line 183: Can we fix this?

rockspore: Yup. Will do

server/config.go
```diff
@@ -216,6 +231,15 @@ func (c *Config) Load(configFile, policySecretPath string) error {
 				return err
 			}
 		}
+
+		// attempts to load the service account credentials if the credentials are not set yet
+		if analyticsSecretPath != "" && c.Analytics.CredentialsJSON == nil {
+			c.Analytics.CredentialsJSON, err = ioutil.ReadFile(path.Join(analyticsSecretPath, ServiceAccount))
+			if err != nil { // not returning the error since it may fall back to using fluentd endpoint
+				c.Analytics.CredentialsJSON = nil
+				log.Warnf("reading analytics service account credentials: %v", err)
```
rockspore, line 240: I used Warnf() on error reading the credentials instead of returning the error because I would like it to fall back to the fluentd endpoint if possible. Let me know if you think a different log level is more appropriate. @theganyo

theganyo: Ah. What's the use case for the fallback? Is it a config mistake? Or could it be intentional?

rockspore: I was considering it as a recovery attempt from a mistake, which could be a config error or k8s secret not applied or mounted correctly.

theganyo: If it's a config mistake, I believe we should error out as the system would not be operating as intended by the config settings.

server/authmanager.go
```diff
@@ -227,3 +227,13 @@ func AuthorizationRoundTripper(config *Config, next http.RoundTripper) (http.Rou
 		return next.RoundTrip(r)
 	}), nil
 }
+
+// NoAuthPUTRoundTripper strips the authorization header for any PUT request
```
theganyo, line 231: The what is clear in the naming and code, it would be more useful for the comment to explain why.

rockspore: Hope it self-explains better now.

theganyo (approved):

PR #60, opened 2020-09-09
Setting up distributed load testing on GKE

### LINKED ISSUE #15: create scripts to setup and run load test

rockspore:
Aimed to fix #15 

Right now one will need to start the load testing through a UI. Needs to have a script to kickstart.


rockspore:
It should now have a self-contained tutorial to set up the distributed load tests from scratch.

theganyo:
Would it make sense to generate `apigee-envoy-adapter.yaml` and `envoy-config.yaml` with the CLI?

rockspore:
> Would it make sense to generate `apigee-envoy-adapter.yaml` and `envoy-config.yaml` with the CLI?

I don't think it is necessary in this particular case since the naming is fairly opinionated and not up for customization. In fact it's also not convenient because debug logging is turned off here but we generally leave it on in simple examples.

theganyo (approved):

PR #62, opened 2020-09-14
use Clone() instead of new transport to ensure intact defaults

theganyo:


server/handler.go
```diff
@@ -100,19 +98,9 @@ func NewHandler(config *Config) (*Handler, error) {
...
-				KeepAlive: 30 * time.Second,
-				DualStack: true,
-			}).DialContext,
-			MaxIdleConns:          100,
-			IdleConnTimeout:       90 * time.Second,
-			TLSHandshakeTimeout:   10 * time.Second,
-			ExpectContinueTimeout: 1 * time.Second,
-			TLSClientConfig:       &tls.Config{InsecureSkipVerify: true},
-		}
+		trans := tr.(*http.Transport).Clone()
```
rockspore, line 101: If my understanding is correct, `Clone()` is to avoid tampering with the `DefaultTransport`. Why not overwrite `tr` directly by?
```
tr = tr.(*http.Transport).Clone()
tr.TLSClientConfig.InsecureSkipVerify = true
```

theganyo: Because `http.DefaultTransport` is actually an `http.RoundTripper` type - which doesn't have access to the TLSClientConfig.

rockspore: Ah I see. Good to know!

rockspore (approved):

PR #64, opened 2020-09-14
add clean shutdown hook

### LINKED ISSUE #63: add clean shutdown hook
Add signal hooks for clean termination from keyboard (SIGINT) and kubernetes (SIGTERM).

theganyo:
closes #63

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,9 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/apigee/apigee-remote-service-envoy/server"
 	"github.com/apigee/apigee-remote-service-golib/log"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -219,4 +222,28 @@ func serve(config *server.Config) {
...
+		sig := <-sigint
+		log.Infof("shutdown signal: %s", sig)
+		signal.Stop(sigint)
+
+		grpcServer.GracefulStop()
+
+		timeout, cancel := context.WithTimeout(context.Background(), time.Second)
+		if err := httpServer.Shutdown(timeout); err != nil {
+			log.Errorf("http shutdown: %v", err)
+			os.Exit(1)
```
rockspore, line 240: It seems to me that if the http server fails to shut down timely, the Close() method will not be called which might leave some staged files. Is that expected?

theganyo: No, good catch... I had meant to remove that Exit().

rockspore (approved):

PR #65, opened 2020-09-17
products must load before /healthz ready

### LINKED ISSUE #52: health check should check products
Currently, the health check returns ok as soon as it starts listening. Instead, it should wait for products to have been loaded at least once.

theganyo:
closes #52

### DIFF
--- a/server/kube_health_test.go
+++ b/server/kube_health_test.go
@@ -1,0 +1,98 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


server/kube_health_test.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	defer productMan.Close()
+
+	grpcHealth := health.NewServer()
+	handler := &Handler{
+		productMan: productMan,
+	}
+
+	kubeHealth := NewKubeHealth(handler, grpcHealth)
+	err = kubeHealth.error()
```
rockspore, line 64: If I didn't miss anything, this is not guaranteed to be called before the product is loaded. While it probably works in normal test condition, it theoretically could fail? Would it make sense to block the product loading in a more robust way till the error is unwanted?

theganyo: Sure. I got lazy. I'll guarantee it fails. (That said, guaranteeing the next test succeeds would be harder.)

rockspore (approved):

PR #69, opened 2020-09-28
Minor fix in config.Load()

rockspore:
This mainly affected the CLI tool by printing extraneous stuff.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -203,10 +203,6 @@ func (c *Config) Load(configFile, policySecretPath, analyticsSecretPath string) 
 		}
 	}
 
-	if err = c.Validate(); err != nil {
-		return err
-	}
-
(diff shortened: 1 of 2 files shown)


rockspore:
I removed the call to [`Validate()`](https://github.com/apigee/apigee-remote-service-envoy/blob/d5c45b50f820d955f513c38ff6aba407913ef4a3/server/config.go#L207
) before the gcp-related secrets are loaded.

This was not causing any error till I edited the config.yaml a bit because by default the config.yaml generated by the CLI contained the fluentd endpoint whether the service account was provided or not. This validation does not seem necessary to me, but let me know if I missed anything. @theganyo

rockspore:
Added tests on loading analytics secret

theganyo (approved):

PR #82, opened 2020-10-14
added timeout for access log streams

### LINKED ISSUE #77: Perf test with the optimized StreamAccessLog service
It'll be worth noting how Envoy balances the load the streaming calls between multiple adapter instances.

rockspore:
It is using the config option to set the timeout for access log streams.

The default duration is changed from 10 minutes to 1 minute. This makes sense to me since terminating connection sooner does little harm when the traffic is low while it enables load balancing more efficiently when traffic is high.

-golib dependencies is changed to the latest.

close #77

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/apigee/apigee-remote-service-envoy
 go 1.15
 
 require (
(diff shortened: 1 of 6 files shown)


rockspore:
closes #77

theganyo (approved):

PR #101, opened 2020-11-13
removed limit on number of resources read from config

### LINKED ISSUE #100: Remove the hard-coded limit (3) of yaml resources that can be read from the config file
This becomes necessary with the introduction of the following issue in the -cli:

https://github.com/apigee/apigee-remote-service-cli/issues/99

rockspore:
close #100


rockspore:
Also modified the integration build to make the k8s sa works with GKE WI.

theganyo (approved):

PR #133, opened 2021-01-07
fixed the env var exports to the sub-shell scripts

rockspore:
This is to fix the kokoro build scripts. The CLI was build and exported in `init.sh`, so its path needs to be retrievable from the subsequent script.


rockspore:
Actually, since the work of moving the build configs internally, this is not needed in the long term and I am closing the PR.

PR #139, opened 2021-01-22
update docker build

theganyo:
updated goboring version
distroless/static for default (13m -> 6m)
distroless/base for goboring (50m -> 14m, more secure)


rockspore:
Curious, what was the problem with the run container of goboring build before?

rockspore (approved):

theganyo:
There was a missing file that the binary was looking for when starting up. Somewhere along the line a new version of either goboring or distroless/base addressed the issue.

PR #140, opened 2021-01-27
multitenant env support

theganyo:
accepts `"*"` as `tenant.env_name` to put it multitenant mode
requires change: https://github.com/apigee/apigee-remote-service-cli/issues/149

For multitenant, Envoy configuration requires the addition of a segment like this to Envoy's `virtual_host` or `route` to specify an environment for ext_authz:
```
                typed_per_filter_config:
                  envoy.filters.http.ext_authz:
                    "@type": type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthzPerRoute
                    check_settings:
                      context_extensions:
                        apigee_environment: test
```

### DIFF
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -62,9 +62,10 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *envoy_auth.Check
 				env,
 			}
 		} else {
+			tracker := prometheusRequestTracker(rootContext)
+			defer tracker.record()
 			err := fmt.Errorf("Envoy must be configured to send Apigee env in ContextExtensions[%s] in multitenant mode.", envContextKey)
-			log.Errorf(err.Error())
-			return nil, err
+			return a.internalError(tracker, err), nil
 		}
 	}
 
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -15,18 +15,18 @@
 package server
 
 import (
-	"context"
+	gocontext "context"
 	"fmt"
 	"net/url"
 	"strings"
 	"time"
 
-	aauth "github.com/apigee/apigee-remote-service-golib/auth"
-	libAuth "github.com/apigee/apigee-remote-service-golib/auth"
+	"github.com/apigee/apigee-remote-service-golib/auth"
+	"github.com/apigee/apigee-remote-service-golib/context"
 	"github.com/apigee/apigee-remote-service-golib/log"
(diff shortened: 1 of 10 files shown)


rockspore:
So it seems possible to enable ext_authz only for specific routes?

server/authorization.go
```diff
@@ -46,14 +47,28 @@ type AuthorizationServer struct {
...
+	if a.handler.isMultitenant {
+		if env, override := req.Attributes.ContextExtensions[envContextKey]; override {
+			rootContext = &multitenantContext{
+				a.handler,
+				env,
+			}
+		} else {
+			err := fmt.Errorf("Envoy must be configured to send Apigee env in ContextExtensions[%s] in multitenant mode.", envContextKey)
+			log.Errorf(err.Error())
+			return nil, err
```
rockspore, line 67: What would Envoy do upon receiving a nil response? Would it make sense to send a denial?

theganyo:
> So it seems possible to enable ext_authz only for specific routes?

Yes. I didn't try that, but it seems like it would be trivial to configure Envoy for that.

server/authorization_test.go
```diff
@@ -237,10 +249,27 @@ func TestCheck(t *testing.T) {
 	if resp.Status.Code != int32(rpc.OK) {
 		t.Errorf("got: %d, want: %d", resp.Status.Code, int32(rpc.OK))
 	}
+
+	// multitenant missing context
+	server.handler.isMultitenant = true
+	if resp, err = server.Check(context.Background(), req); err == nil {
```
rockspore, line 255: This `resp` is unused and should be replaced with `_`.

rockspore (changes requested):

server/header_context_test.go
```diff
@@ -80,19 +84,29 @@ func TestMetadataHeadersExceptions(t *testing.T) {
...
+		envName: "*",
 	}
 	h.targetHeader = "target"
-	header := map[string]string{"target": "target"}
+	header := map[string]string{
+		"target":          "target",
+		headerEnvironment: "test",
+	}
+
 	api, ac := h.decodeMetadataHeaders(header)
```
rockspore, line 95: nit: this `api` assignment is not ineffectual and should be `_`. golangci actually spotted this but had it filtered out.

rockspore (approved):

PR #148, opened 2021-02-10
ext_authz service emits metadata instead of headers for subsequent RBAC and ALS

### LINKED ISSUE #146: use metadata instead of headers for authz -> ALS
Currently, Apigee auth metadata is communicated to the access log service [via headers](https://github.com/apigee/apigee-remote-service-envoy/blob/master/server/header_context.go#L26). This was the only mechanism to do so in the Envoy v2 API. However, unless these headers are then stripped by properly configuring another Envoy filter, they will be sent on to the target service.

Envoy has added the ability to send this information using dynamic metadata in the [v3 API](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#service-auth-v3-checkresponse). Envoy Adapter should use this mechanism instead of headers. 

Requirements
- Switch to dynamic metadata instead of headers in the ext_authz response
- Only set headers if append metadata headers is set
- Use dynamic metadata instead of headers for RBAC
- Use dynamic metadata for analytics

rockspore:
This avoids forwarding those headers to the target service without configuring an extra header_to_metadata filter.

close #146


rockspore:
Summary:

A boolean config option `server.Config.Auth.AppendMetadataHeaders` is added, which will be default to true in the CLI during provisioning.
* This lets the ext_authz.Check service append the request headers in addition to sending dynamic metadata.
* It also informs the StreamAccessLog service whether to fall back to check headers for authContext and target.

theganyo (approved):

PR #156, opened 2021-02-16
reject_unauthorized changed to allow_unauthorized

theganyo:


rockspore (approved):
Thanks for fixing the `emit_metadata_headers` field!

PR #169, opened 2021-03-11
target replaced with api

### LINKED ISSUE #168: Replace `target` with `api` or something similar and more appropriate
Specifically, this is to replace the metadata key `apigee_target` with `apigee_api`, `targetHeader` in the config with `apiHeader`. Meanwhile, anywhere `target` is used as a function argument will be replaced with `api` or something similar.

Reasoning is that, although `target` is by default the host header of a request, it is more aligned with the name of a `api_proxy` in Apigee. Moving away from `target` makes it less confusing, especially when `target` is overloaded elsewhere.

rockspore:
close #168

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.16
 // replace github.com/apigee/apigee-remote-service-golib => ../apigee-remote-service-golib
 
 require (
-	github.com/apigee/apigee-remote-service-golib v1.4.1-0.20210224224933-6a2cf53cd4b7
+	github.com/apigee/apigee-remote-service-golib v1.4.1-0.20210312005113-b7a239f20866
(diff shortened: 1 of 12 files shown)


theganyo (dismissed):
Looks good. You could update the go.mod in this PR once you've addressed https://github.com/apigee/apigee-remote-service-golib/issues/55... or just leave it until we tag a version. Up to you.

rockspore:
> Looks good. You could update the go.mod in this PR once you've addressed [apigee/apigee-remote-service-golib#55](https://github.com/apigee/apigee-remote-service-golib/issues/55)... or just leave it until we tag a version. Up to you.

Makes sense. I will hold on the merge and wait for -golib. Meanwhile, there may be some change needed in -cli. I will take a look.

theganyo (approved):

PR #172, opened 2021-03-16
added mtls support

rockspore:
- OPDK only for analytics
- All Apigee platforms for other remote-service endpoints

Analytics mTLS is configured in `server.Config.Analytics.TLS`.
Others is configured in `server.Config.Tenant.TLS` - `AllowUnverifiedSSLCert` is also moved inside.

### DIFF
--- a/server/config.go
+++ b/server/config.go
@@ -116,19 +116,19 @@ type TLSClientConfig struct {
 
 // TenantConfig is config relating to an Apigee tentant
 type TenantConfig struct {
-	InternalAPI            string          `yaml:"internal_api,omitempty" json:"internal_api,omitempty"`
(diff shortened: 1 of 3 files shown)


theganyo:
It looks like we could easily get better test coverage by passing in bad file names in the config and checking for those errors.

theganyo:
Also, we can cover roundTripperWithTLS() with it's own test(s).

rockspore:
> Also, we can cover roundTripperWithTLS() with it's own test(s).

Forgot this possibility. Yes will do!

rockspore:
In addition to just covering the errors, added one test to verify the client actually sends the cert to the test server.

theganyo (approved):

theganyo:
Please be sure to update Will.

PR #196, opened 2021-04-06
Define EnvironmentConfig

### LINKED ISSUE #195: define operations configurations
Issue to track the first task of #189

rockspore:
This only contains definition of related go structs. Loading and validation functions will come in a separate PR once the definition is settled.

Close #195


arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package arcconfig defines the API Runtime Control config and provides
+// the config loading and validation functions.
+package arcconfig
```
josheinhorn, line 17: Is `config` already taken? The `arc` prefix does not seem to add any more specificity

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+
+// APIRuntimeControlConfig is an Apigee Environment-level config for
+// Envoy Adapter. It contains a list of operations for the adapter to
+// perform request authentication and authorization.
+type APIRuntimeControlConfig struct {
+	// Name of the config
+	Name string `yaml:"name"`
+
+	// Revision of the config
+	Revision string `yaml:"revision"`
```
josheinhorn, line 29: Can we use `ID` (it does not imply order or versioning)?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+// JWTClaim has reference to a JWT claim.
+type JWTClaim struct {
+	// Name of the JWT provider
+	Provider string `yaml:"provider"`
+
+	// Name of the claim
+	Name string `yaml:"name"`
+}
+
+type HTTPMethod int
```
theganyo, line 151: This will be difficult for human's to read in the config. Perhaps we should use standard http strings instead of arbitrary ints. Doing so would also map to Go's standard http.* method constants.

rockspore: That's true. I forgot that it should also be more human writable. Changed to string and we will validate it upon loading.

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,149 @@
...
+
+// APIRuntimeControlConfig is an Apigee Environment-level config for
+// Envoy Adapter. It contains a list of operations for the adapter to
+// perform request authentication and authorization.
+type APIRuntimeControlConfig struct {
+	// Name of the config
+	Name string `yaml:"name"`
+
+	// Revision of the config
+	Revision string `yaml:"revision"`
```
theganyo, line 29: Are we using the name or revision or both for matching? We need to document what these mean wrt. our logic.

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+package arcconfig
+
+import "time"
+
+// APIRuntimeControlConfig is an Apigee Environment-level config for
+// Envoy Adapter. It contains a list of operations for the adapter to
+// perform request authentication and authorization.
+type APIRuntimeControlConfig struct {
+	// Name of the config
+	Name string `yaml:"name"`
```
josheinhorn, line 26: What is this for and how is it different from ID (Revision)?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+// Package arcconfig defines the API Runtime Control config and provides
+// the config loading and validation functions.
+package arcconfig
+
+import "time"
+
+// APIRuntimeControlConfig is an Apigee Environment-level config for
+// Envoy Adapter. It contains a list of operations for the adapter to
+// perform request authentication and authorization.
+type APIRuntimeControlConfig struct {
```
josheinhorn, line 24: We're missing a layer of indirection, we need the top level (the Environment), and then another layer for each Proxy (i.e. per base path) so something like:

```
type EnvironmentConfig struct {
  ID string
  []ProxyConfig ProxyConfigs
}

type ProxyConfig {
  BasePath string
  []Operation Operations 
  ...
}
```

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// An API Operation associates a set of rules with a set of request matching
+// settings.
+type Operation struct {
+	// Name of the operation. Unique within a APIRuntimeControlConfig.
+	Name string `yaml:"name"`
+
+	// Authentication defines the AuthenticationRequirement
+	Authentication AuthenticationRequirement `yaml:"authentication,omitempty"`
```
josheinhorn, line 42: Just checking; does this inherently support JSON?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+// perform request authentication and authorization.
+type APIRuntimeControlConfig struct {
+	// Name of the config
+	Name string `yaml:"name"`
+
+	// Revision of the config
+	Revision string `yaml:"revision"`
+
+	// A list of Operations, names of which must be unique within the config.
+	Operations []*Operation `yaml:"operations,omitempty"`
```
josheinhorn, line 32: Nit: pointer is not necessary as there are no methods on the `Operation` struct and memory optimization is premature

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+	In []HTTPParameter `yaml:"in"`
+}
+
+// RemoteJWKS contains information for remote JWKS
+type RemoteJWKS struct {
+	// URL of the JWKS
+	URL string `yaml:"url"`
+
+	// CacheDuration of the JWKS
+	CacheDuration time.Duration `yaml:"cache_duration,omitempty"`
```
josheinhorn, line 100: What does the YAML form of this look like?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+
+// HTTPMatch is an HTTP request matching rule
+type HTTPMatch struct {
+	// URL path template using to match incoming requests and optionally identify
+	// path variables. This should be fully denormalized as there is no global
+	// basepath defined anywhere.
+	PathTemplate string `yaml:"path_template"`
+
+	// HTTP method
+	Method HTTPMethod `yaml:"method,omitempty"`
```
josheinhorn: I see this is an `int` alias, however the YAML form will be a string e.g. `GET`, `POST`, etc. How do we plan on reconciling this?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+
+// ConsumerAuthorization is the configuration of API consumer authorization
+type ConsumerAuthorization struct {
+	FailOpen bool `yaml:"fail_open,omitempty"`
+
+	// Locations of API consumer credential (API Key). First match wins.
+	In []HTTPParameter `yaml:"in"`
+
+	// Quota identifier (optional)
+	QuotaIdentifier string `yaml:"quota_identifier,omitempty"`
```
josheinhorn, line 111: Should this actually be in scope for the first pass @theganyo ? It requires some amount of changes to ARC's quota impl so unless we have a reason we can't add this in a non-breaking manner later, I recommend we remove it for now.

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+	Name string `yaml:"name"`
+
+	// Authentication defines the AuthenticationRequirement
+	Authentication AuthenticationRequirement `yaml:"authentication,omitempty"`
+
+	// Authorization defines the ConsumerAuthorization
+	Authorization ConsumerAuthorization `yaml:"authorization,omitempty"`
+
+	// HTTP matching rules for this operation.
+	// The path templates within which should be fully denormalized.
```
josheinhorn, line 48: Is this implying it includes the basepath? We decided last week that we'd keep basepath at a higher level i.e. the first layer of the tree.

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,164 @@
...
+// settings.
+type Operation struct {
+	// Name of the operation. Unique within a APIRuntimeControlConfig.
+	Name string `yaml:"name"`
+
+	// Authentication defines the AuthenticationRequirement
+	Authentication AuthenticationRequirement `yaml:"authentication,omitempty"`
+
+	// Authorization defines the ConsumerAuthorization
+	Authorization ConsumerAuthorization `yaml:"authorization,omitempty"`
```
josheinhorn, line 45: ConsumerAuthorization (need to differentiate from other kinds of Authz we may add in the future).

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty"`
+
+	// Prefix to strip off matched value (e.g. "Bearer " for Authorization
+	// tokens).
+	Prefix string `yaml:"prefix,omitempty"`
+}
+
+// JWTClaim has reference to a JWT claim.
```
josheinhorn, line 142: is

josheinhorn (changes requested):

arcconfig/arcconfig.go
theganyo: Let's remove.

arcconfig/arcconfig.go
rockspore: I actually just changed this to string as Scott had the similar comment.

arcconfig/arcconfig.go
rockspore: Yes I think the basepath should be kept at a higher level for the other schema. I just thought when we generate this config from it, we could denormalize all the paths here so that ARC/EA doesn't have to compute this. Also we need to do the denormalization in control plane anyway for the ingress contract generation.

arcconfig/arcconfig.go
rockspore: Possibly not. But it's easy to add the json annotation along with the yaml one. Will do it.

arcconfig/arcconfig.go
theganyo: Yeah... probably we should keep this at the higher level. That matches Open API's approach as well. Computing the full path is trivial, but requiring users to enter it repeatedly is lame. :)

arcconfig/arcconfig.go
theganyo: I think Go requires both annotations to support both... if we are.

arcconfig/arcconfig.go
rockspore: `config` is still available. There is a config.go file in the `server` pkg though. FYI, I made this separate package because I thought it would be simpler to import it.

arcconfig/arcconfig.go
theganyo: Yup. I just asked about this as well. We do need to tie these to an environment somehow.

arcconfig/arcconfig.go
theganyo: Yup. We decided to remove and just use ID.

rockspore: Will remove it and just use ID.

arcconfig/arcconfig.go
rockspore: Makes sense. Will add a top one.

arcconfig/arcconfig.go
rockspore: As replied in other comments. Will use ID only and require its uniqueness.

arcconfig/arcconfig.go
theganyo: We may want to refactor/rename a little. We have a global "config" in the server package.

arcconfig/arcconfig.go
rockspore: re: Scott - I just thought we still need the original config file anyway which indicates the environment. As such, we only need a list of these configs assuming they are all tied to the specified environment. Or are we ditching the original config entirely if we see any of these ones?

Will add this layer. Apart from this, my idea was to have each operation self-contained here instead of having those global settings. I don't have strong opinion. If the goal is to make the structures here as alike the schema as possible, I will add those global rules back.

theganyo: Yeah, this is a little tricky.... Right now we have a global config element for env. And we have the ability to set the env from Envoy via metadata. We could keep using that scheme even with this config file. But only if these configs are not required to be tied to a specific env?

arcconfig/arcconfig.go
rockspore: Ack.

arcconfig/arcconfig.go
rockspore: Ack.

arcconfig/arcconfig.go
rockspore: For example, `30s` or `1h`.

arcconfig/arcconfig.go
rockspore: Done.

arcconfig/arcconfig.go
josheinhorn: So the requirement here is simply that there is some unique identifier per config (the `ID` field) that is linked to a complete group of Proxy configurations (basepath + operations). I don't fully understand how this will fit into the other "global" element... maybe a small YAML example would clarify?

arcconfig/arcconfig.go
josheinhorn: I recommend we use JSON when a machine generates this config, so we should at some point add `json` annotations.

arcconfig/arcconfig.go
josheinhorn: I see... this is fine the way it is then, we can always refactor it later without breaking the contract with clients.

arcconfig/arcconfig.go
rockspore: Added in the latest comment.

arcconfig/arcconfig.go
josheinhorn: Actually reflecting on this, I recall a discussion that the global config would contain *references* to other files, where each file was a single `EnvironmentConfig` (i.e. group of Proxies).

rockspore: In case it helps, the existing "global" config looks like this:
```
global:
  temp_dir: /tmp/apigee-istio
  keep_alive_max_connection_age: 10m
  api_address: :5000
  metrics_address: :5001
tenant:
  remote_service_api: https://org-test.apigee.net/remote-service
  org_name: org
  env_name: env
  key: mykey
  secret: mysecret
  client_timeout: 30s
products:
  refresh_rate: 2m
analytics:
  file_limit: 1024
  send_channel_size: 10
  collection_interval: 10s
auth:
  api_key_claim: claim
  api_key_cache_duration: 30m
```
My understanding is if the configs defined here are specified, then the `auth` section in the original one will be ignored.

josheinhorn: Can we add a new field? e.g. 

```
arc_env_configs: ["file1", "file2"]
```

This would be mutually exclusive with `auth`

rockspore: The idea was to take these new configs via a new command line flag. But this new field makes sense to me as it gives the feasibility to still keep everything in one file.

theganyo: Yes, we could add a section like this. My original thought was that we'd simply accept the arc configs as a CLI params... but this would be more formal.

arcconfig/arcconfig.go
theganyo: FYI, this is the standard for Go. https://pkg.go.dev/maze.io/x/duration#ParseDuration

arcconfig/arcconfig.go
rockspore: It seems we all agree on such a new section then. Should I move everything into the `server` pkg then?

theganyo: Perhaps we should relocate all the config structs and processing to a config package.

rockspore: That sounds better to me. It wouldn't break any user-facing interfaces.

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,150 @@
...
+// ConsumerAuthorization is the configuration of API consumer authorization
+type ConsumerAuthorization struct {
+	FailOpen bool `yaml:"fail_open,omitempty" json:"fail_open,omitempty"`
+
+	// Locations of API consumer credential (API Key). First match wins.
+	In []HTTPParameter `yaml:"in" json:"in"`
+}
+
+// HTTPMatch is an HTTP request matching rule
+type HTTPMatch struct {
```
josheinhorn, line 117: What about `path_rewrite`? This can be done via Ext Authz IIUC via the `:path` pseudo-header, right?

arcconfig/arcconfig.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// Name of a header
+	Header string `yaml:"header,omitempty" json:"header,omitempty"`
+
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty" json:"jwt_claim,omitempty"`
+
+	// Prefix to strip off matched value (e.g. "Bearer " for Authorization
+	// tokens).
+	Prefix string `yaml:"prefix,omitempty" json:"prefix,omitempty"`
```
josheinhorn, line 140: Should we just go straight to a holistic substitution implementation e.g. 
```
template: "prefix-{foo}-{bar}-suffix"
substitution: "{foo}_{bar}"
```
-->
```
input: "prefix-hello-world-suffx"
output: "hello_world"
```

This covers the prefix case and doesn't paint us into a corner. Of course we should add limits, both to string lengths and # of template variables.

arcconfig/arcconfig.go
rockspore: Can ext_authz rewrite headers?

rockspore:
I refactored the configs into a `config` package. So the PR has become a little sizable. Primary change is the added code in the `./config/config.go`. @theganyo @josheinhorn

arcconfig/arcconfig.go
rockspore: I changed it to StringSubstitution with a definition identical to the schema.

theganyo: Are we not using mustache-style templates (ie. `{{` and `}}`)?

rockspore: Yeah we should do `{{}}`. Modified the comment.

arcconfig/arcconfig.go
theganyo: We can certainly send override headers, so it should be possible.

theganyo: https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-msg-service-auth-v3-okhttpresponse

rockspore: Ah, I see. So it is able to rewrite the header. What about reusing the StringModification here instead of defining two strings `path_rewrite` and `path_template`?

theganyo: Actually, I feel like if we're doing request modification that it should be a separate section...

rockspore: Yeah I have the same feeling. It should be another field in `Operation`?

theganyo: Yes. Maybe Target should be a section that contains request rewrite information?

config/config.go
```diff
@@ -161,6 +166,160 @@ type AuthConfig struct {
...
+
+	// Name of a header
+	Header string `yaml:"header,omitempty" json:"header,omitempty"`
+
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty" json:"jwt_claim,omitempty"`
+
+	// String modification to strip off matched value (e.g. "Bearer " for Authorization
+	// tokens).
+	Modification StringModification `yaml:"modification,omitempty" json:"modification,omitempty"`
```
josheinhorn, line 297: Nit: I'm flip-flopping between `Modification` and `Transformation`?

rockspore: `Transformation` sounds cooler to me :)

theganyo: +1 Transformation

arcconfig/arcconfig.go
theganyo: Based on other thread's naming suggestion, perhaps a "request_transformation" section?

rockspore: The name LGTM..but put `Target` under such a name seems a bit weird..

theganyo: No, I wouldn't put Target under that... I was thinking that Target would be a section that has a transformations section.

rockspore: I put a `RequestTransformation` under `Target`. It also feels like it could be put directly under `Operation`. I would be fine with either way.

config/config.go
rockspore: Ok. So we have the majority :)

util/util.go
```diff
@@ -90,3 +98,29 @@ func WriteProperties(writer io.Writer, props map[string]string) error {
 
 	return nil
 }
+
+// LoadPrivateKey load private key bytes into rsa.PrivateKey
+func LoadPrivateKey(privateKeyBytes []byte) (*rsa.PrivateKey, error) {
```
theganyo, line 103: How did we lose coverage of this function?

rockspore: This was indirectly covered in the `server` package. There was no tests for it. I am aware of it and planning to add tests for it in a separate PR.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty" json:"jwt_claim,omitempty"`
+
+	// String modification to strip off matched value (e.g. "Bearer " for Authorization
+	// tokens).
+	Transformation StringTransformation `yaml:"transformation,omitempty" json:"transformation,omitempty"`
+}
+
+// StringTransformation uses simple template syntax
+// e.g. template: "prefix-{{foo}}-{{bar}}-suffix"
```
josheinhorn, line 315: Let's use the Google standard for request matching syntax (single curly braces) unless we have reason to diverge.

This also directly collides with the upcoming Environment variable templating, which will use mustache-like syntax (double curly brace).

josheinhorn (changes requested):

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
 	AppendMetadataHeaders bool          `yaml:"append_metadata_headers,omitempty" json:"append_metadata_headers,omitempty"`
 }
 
+// EnvConfigs
+type EnvConfigs struct {
+	// A list of files containing environment config
+	ConfigFiles []string `yaml:"config_files,omitempty" json:"config_files,omitempty"`
```
josheinhorn, line 172: Do we want to actually prescribe that these must be "files", or should we be more generic and support URIs, where files can be referencing using the standard `file://` scheme. So in this case we might call this `ConfigURIs` or `ConfigRefs`

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	Any []AuthenticationRequirement `yaml:"any,omitempty" json:"any,omitempty"`
+
+	// All contains a list of AuthenticationRequirements.
+	// A successful authentication requires all of them being satisfied.
+	All []AuthenticationRequirement `yaml:"all,omitempty" json:"all,omitempty"`
+}
+
+// JWTAuthentication defines the JWT authentication
+type JWTAuthentication struct {
+	// Name of this JWT Provider, unique within the Proxy.
```
josheinhorn, line 236: This feels more like a "Requirement" than a "Provider". I bring this up because we use the term `Provider` below in `JWTClaim` which looks odd when referring to this object.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	Name string `yaml:"name" json:"name"`
+
+	// JWT issuer ("iss" claim)
+	Issuer string `yaml:"issuer" json:"issuer"`
+
+	// A remote JWKS source
+	RemoteJWKS RemoteJWKS `yaml:"remote_jwks" json:"remote_jwks"`
```
josheinhorn, line 243: Question: how do we plan on introducing alternative JWKS sources i.e. a oneof?

arcconfig/arcconfig.go
josheinhorn: I agree it's not strictly a matching rule and is really a "transform." However, the problem with isolating the transform elsewhere is that `HTTPMatch` defines the _template_, so duplicating the template elsewhere can be error-prone and  confusing. Note that Envoy colocates the match rule and rewrite rule in the route config. Let's move this out of scope given the lack of clarity around the requirements.

Also let's definitely _not_ put it in `Target` when we do add it. Target is akin to a Cluster, while matching/transforming is correlated to Filter config (i.e. Router or HCM filters)

config/config.go
rockspore: Yeah I realized the collision this morning too, which would prevent us from conveniently using the template library there.

rockspore: Will change it back to single curly braces.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
 	AppendMetadataHeaders bool          `yaml:"append_metadata_headers,omitempty" json:"append_metadata_headers,omitempty"`
 }
 
+// EnvConfigs
+type EnvConfigs struct {
+	// A list of files containing environment config
+	ConfigFiles []string `yaml:"config_files,omitempty" json:"config_files,omitempty"`
+
+	// A list of environment configs
+	InlineConfigs []EnvironmentConfig `yaml:"inline_configs,omitempty" json:"inline_configs,omitempty"`
```
josheinhorn, line 175: Nit: just `Inline` is sufficiently specific.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	// Top-level basepath for the proxy config
+	Basepath string `yaml:"basepath,omitempty" json:"basepath,omitempty"`
+
+	// A list of Operations, names of which must be unique within the proxy config.
+	Operations []Operation `yaml:"operations,omitempty" json:"operations,omitempty"`
+}
+
+// An API Operation associates a set of rules with a set of request matching
+// settings.
+type Operation struct {
```
josheinhorn, line 200: Will this possibly collide with the API Product Operation definition?

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	In []HTTPParameter `yaml:"in" json:"in"`
+}
+
+// Target contains the name of the target server and request transformation
+type Target struct {
+	// Name of the target
+	Name string `yaml:"name" json:"name"`
+
+	// RequestTransformation defines how the request should be modified, e.g. a path rewrite
+	RequestTransformation RequestTransformation `yaml:"request_transformation,omitempty" json:"request_transformation,omitempty"`
```
josheinhorn: Let's remove this until we have clear requirements.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+}
+
+type RequestTransformation struct {
+	// StringTransformation to rewrite the path request header
+	PathRewrite StringTransformation `yaml:"path_rewrite,omitempty" json:"path_rewrite,omitempty"`
```
josheinhorn: FWIW, this is error-prone; this re-defines a `template` but the preceding `HTTPMatch` already defines the `template`. As mentioned above, let's remove this until we have clarity.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	// ConsumerAuthorization defines the ConsumerAuthorization
+	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
+
+	// HTTP matching rules for this operation.
+	HTTPMatches []HTTPMatch `yaml:"http_match,omitempty" json:"http_match,omitempty"`
+
+	// Target server for this operation. This will be sent to Envoy
+	// for routing to the corresponding upstream cluster upon a successful
+	// authorization of the operation.
+	Target Target `yaml:"target" json:"target"`
```
josheinhorn: This should just be a `string`, anything else is a smell that we're doing too much routing work in ARC

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+	// Name of a query paramter
+	Query string `yaml:"query,omitempty" json:"query,omitempty"`
+
+	// Name of a header
+	Header string `yaml:"header,omitempty" json:"header,omitempty"`
+
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty" json:"jwt_claim,omitempty"`
+
+	// String modification to strip off matched value (e.g. "Bearer " for Authorization
```
josheinhorn, line 301: Just "Optional transformation of the parameter value."

config/config.go
rockspore: They are in different packages so technically no. But I have changed it to `APIOperation`.

config/config.go
rockspore: Done.

config/config.go
rockspore: I defined a new struct `JWKSSource` and put `RemoteJWKS` under it. So when we introduce more sources, we can enforce the "oneof" requirement in the parent struct.

config/config.go
rockspore: Changed it to requirement.

config/config.go
rockspore: Changed to `ConfigURIs` and added a comment that only local file system is supported currently.

config/config.go
```diff
@@ -161,6 +166,174 @@ type AuthConfig struct {
...
+// Precisely one of Query, Header or JWTClaim should be set.
+type HTTPParameter struct {
+	// Name of a query paramter
+	Query string `yaml:"query,omitempty" json:"query,omitempty"`
+
+	// Name of a header
+	Header string `yaml:"header,omitempty" json:"header,omitempty"`
+
+	// A JWTClaim
+	JWTClaim JWTClaim `yaml:"jwt_claim,omitempty" json:"jwt_claim,omitempty"`
```
josheinhorn, line 299: These three fields are mutually exclusive, are we planning on writing validation for this? JFYI there is an alternative which is the use pseudo-union type and a custom YAML/JSON marshaller impl, which makes the usage of the config much more straightforward via type switch e.g.

```
type ParamMatch interface {
  paramMatch()
}

type HTTPParameter struct {
  Match ParamMatch
  Transformation StringTransformation
}

type Query string
func (Query)  paramMatch() {}

type Header string
func (Header)  paramMatch() {}

type JWTClaim struct {...}
func (JWTClaim)  paramMatch() {}

func (p *HTTPParameter) UnmarshalJSON(b []byte) error {
  // implement custom logic for the Match field
}

func exampleUsage() {
  p := &HTTPParameter{}
  _ := json.Unmarshal(p, data)
  switch match := p.Match.(type) {
    case Query:
      ...
    case Header:
      ...
    case JWTClaim:
      ...
}
```

josheinhorn (changes requested):

config/config.go
```diff
@@ -161,6 +166,166 @@ type AuthConfig struct {
...
+// ProxyConfig has the proxy configuration
+type ProxyConfig struct {
+	// Top-level basepath for the proxy config
+	Basepath string `yaml:"basepath,omitempty" json:"basepath,omitempty"`
+
+	// A list of Operations, names of which must be unique within the proxy config.
+	Operations []APIOperation `yaml:"operations,omitempty" json:"operations,omitempty"`
+}
+
+// An APIOperation associates a set of rules with a set of request matching
```
josheinhorn, line 199: Go readability: Comment should start with the struct name (remove "An ")

config/config.go
rockspore: Ah yeah that will be a good idea. The same thing can be done for the `JWKSSource` too.

config/config.go
rockspore: ("An", "A" or "The") is actually allowed by the style decisions. But I realize I am missing a bunch of periods. :)

config/config.go
rockspore: Changed `JWKSSource` to interface implemented by `RmoteJWKS` and potentially other sources in the future.

config/config.go
rockspore: Yes. Changed `Target` back to a string.

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+// RemoteJWKS contains information for remote JWKS.
+type RemoteJWKS struct {
+	// URL of the JWKS
+	URL string `yaml:"url" json:"url"`
+
+	// CacheDuration of the JWKS
+	CacheDuration time.Duration `yaml:"cache_duration,omitempty" json:"cache_duration,omitempty"`
+}
+
+func (RemoteJWKS) jwksSource() {}
```
josheinhorn, line 271: Is there a way to suppress the codecov warnings using comments? Writing a test for a no-op func is clearly silly.

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+
+// Name of a header
+type Header string
+
+func (Header) paramMatch() {}
+
+// JWTClaim is reference to a JWT claim.
+type JWTClaim struct {
+	// Name of the JWT provider
+	Provider string `yaml:"provider" json:"provider"`
```
josheinhorn, line 318: I ended up deciding on `Requirement` for this field in the API schema.

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+type EnvironmentConfig struct {
+	// Unique ID of the environment config
+	ID string `yaml:"id" json:"id"`
+
+	// A list of proxy configs
+	ProxyConfigs []ProxyConfig
+}
+
+// ProxyConfig has the proxy configuration.
+type ProxyConfig struct {
```
josheinhorn, line 191: Are we intentionally omitting the "default" settings (target, authn, authz)?

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+	// Name of the operation. Unique within a APIRuntimeControlConfig.
+	Name string `yaml:"name" json:"name"`
+
+	// Authentication defines the AuthenticationRequirement
+	Authentication AuthenticationRequirement `yaml:"authentication,omitempty" json:"authentication,omitempty"`
+
+	// ConsumerAuthorization defines the ConsumerAuthorization
+	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
+
+	// HTTP matching rules for this operation.
```
josheinhorn, line 211: Note that if omitted, this is supposed to match all operations (similar to the "default" case in a switch)

config/config.go
rockspore: I would like each `APIOperation` to be self-contained here. But it's open for debate.

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+// JWTAuthentication defines the JWT authentication.
+type JWTAuthentication struct {
+	// Name of this JWT requirement, unique within the Proxy.
+	Name string `yaml:"name" json:"name"`
+
+	// JWT issuer ("iss" claim)
+	Issuer string `yaml:"issuer" json:"issuer"`
+
+	// The JWKS source
+	JWKSource JWKSSource
```
josheinhorn, line 244: Typo: JWKSSource

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+
+	// Name of the target server for this operation. This will be sent to Envoy
+	// for routing to the corresponding upstream cluster upon a successful
+	// authorization of the operation.
+	Target string `yaml:"target" json:"target"`
+}
+
+// AuthenticationRequirement defines the authentication requirement.
+// Precisely one of JWT, Any and All should be set.
+type AuthenticationRequirement struct {
```
josheinhorn, line 222: Looks like we're using the pseudo-union type elsewhere for oneofs ... should we do the same here?

config/config.go
```diff
@@ -161,6 +166,177 @@ type AuthConfig struct {
...
+	HTTPMatches []HTTPMatch `yaml:"http_match,omitempty" json:"http_match,omitempty"`
+
+	// Name of the target server for this operation. This will be sent to Envoy
+	// for routing to the corresponding upstream cluster upon a successful
+	// authorization of the operation.
+	Target string `yaml:"target" json:"target"`
+}
+
+// AuthenticationRequirement defines the authentication requirement.
+// Precisely one of JWT, Any and All should be set.
```
rockspore, line 221: Should we also convert this to an interface here or keep it as it and validate it ourselves? @josheinhorn @theganyo

config/config.go
rockspore: I am not sure if I understand this comment. We have to have HTTPMatch right?

config/config.go
rockspore: Per offline discussion. I put the global target, authn and authz.

rockspore:
I am keeping this PR for now, since a decision is yet to made whether things should be defined here. The bottom line is two tentative schemas break at the `Target` definition.

config/config.go
rockspore: Done.

config/config.go
```diff
@@ -161,6 +166,194 @@ type AuthConfig struct {
...
+	// authorization of the operation.
+	Target string `yaml:"target" json:"target"`
+}
+
+// AuthenticationRequirement is the interface defining the authentication requirement.
+type AuthenticationRequirement interface {
+	authenticationRequirement()
+}
+
+// AnyAuthenticationRequirements requires any of emclosed requirements to be satisfied for a successful authentication.
```
theganyo, line 236: :s/emclosed/enclosed

config/config.go
```diff
@@ -161,6 +166,194 @@ type AuthConfig struct {
...
+type AnyAuthenticationRequirements struct {
+	Any []AuthenticationRequirement
+}
+
+func (AnyAuthenticationRequirements) authenticationRequirement() {}
+
+// AllAuthenticationRequirements requires all of emclosed requirements to be satisfied for a successful authentication.
```
theganyo, line 243: :s/emclosed/enclosed

theganyo:
Do we need to add a "RequestTransformations" or similar section to allow for path rewrites?

config/config.go
theganyo: Not that I'm aware of.

rockspore:
> Do we need to add a "RequestTransformations" or similar section to allow for path rewrites?

Yes. Under `APIOperation`? The `path_template` is inside the `http_match` though. Or we rename `http_match`?

theganyo:
I think we'll eventually probably want to support more transformations than just path. Also, it doesn't make a lot of sense to me to put a path transformation in the matcher.

theganyo:
Anyway, Josh said earlier that we'd potentially just move transformations out of scope. If so, the rest LGTM.

theganyo (dismissed):

config/config.go
rockspore: Added tests anyway.

rockspore:
@josheinhorn Is this good to be merged for now?

josheinhorn (approved):

config/config.go
rockspore: Ah, gotcha. Will add a comment.

config/config.go
```diff
@@ -161,6 +166,196 @@ type AuthConfig struct {
...
+	Target string `yaml:"target" json:"target"`
+}
+
+// AuthenticationRequirement is the interface defining the authentication requirement.
+type AuthenticationRequirement interface {
+	authenticationRequirement()
+}
+
+// AnyAuthenticationRequirements requires any of enclosed requirements to be satisfied for a successful authentication.
+type AnyAuthenticationRequirements struct {
```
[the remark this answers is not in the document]
rockspore: Ack.

config/config.go
```diff
@@ -161,6 +166,196 @@ type AuthConfig struct {
...
+	Target string `yaml:"target" json:"target"`
+
+	// A list of Operations, names of which must be unique within the proxy config.
+	Operations []APIOperation `yaml:"operations,omitempty" json:"operations,omitempty"`
+}
+
+// An APIOperation associates a set of rules with a set of request matching
+// settings.
+type APIOperation struct {
+	// Name of the operation. Unique within a APIRuntimeControlConfig.
```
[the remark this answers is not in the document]
rockspore: Yes. Will change it.

config/config.go
```diff
@@ -161,6 +166,196 @@ type AuthConfig struct {
...
+
+	// Authentication defines the proxy-level authentication requirement
+	Authentication AuthenticationRequirement `yaml:"authentication,omitempty" json:"authentication,omitempty"`
+
+	// ConsumerAuthorization defines the proxy-level consumer authorization
+	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
+
+	// Name of the target server for this proxy. This will be sent to Envoy
+	// for routing to the corresponding upstream cluster upon a successful
+	// authorization.
```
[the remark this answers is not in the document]
rockspore: Ack.

josheinhorn (approved):

PR #198, opened 2021-04-07
set up go1.16 before golangci-lint

rockspore:
In the Lint workflow of GitHub Actions, reviewdog has been complaining a few issues that shouldn't be there for go1.16


theganyo (approved):
Does this need to be done in the other repos as well?

rockspore:
> Does this need to be done in the other repos as well?

Ah, that's right. Will do it.

PR #199, opened 2021-04-13
Using environment variables for configuration

### LINKED ISSUE #165: Twelve factor app guidelines: using environment variables for configuration
## What's missing?

The go app currently uses cobra to read a 'config' file for all its configuration settings. This config file contains configuration variables, and secrets, that a user cannot override easily using environment variable(s).

## Your preferred solution

https://12factor.net/config

Using APIGEE_X_Y_Z* environment variables to override settings would simplify deployment. Can the current implementation be extended with [viper](https://github.com/spf13/viper) to support this feature?

With Kubernetes it will also open the possibility to use separate Secret(s) and ConfigMap to store config variables.

rockspore:
Enabled env var binding for loading all configs except for the new envconfig.

Close #165


theganyo:
Let's make sure our loading/resolving mechanism is documented at the top of the file.

config/config.go
```diff
@@ -356,7 +402,29 @@ type StringTransformation struct {
 	Substitution string `yaml:"substitution,omitempty" json:"substitution,omitempty"`
 }
 
-// Load config
+// Load config with the given config file, secret paths and a flag specifying whether analytics credentials must be present.
```
rockspore, line 405: I added a chunk of documentation on `Load()` to describe how the config gets loaded. PTAL @theganyo

theganyo:
General question: Is there a reason to use a combination of underscore and dot notation?

rockspore:
> General question: Is there a reason to use a combination of underscore and dot notation?

Dot indicates separates structs, although this may not be super helpful to users.

theganyo:
Makes sense... and we obviously can't use only underscores since we have some variable names with underscore. Would it make sense to go the other direction and only use dot notation?

rockspore:
> Makes sense... and we obviously can't use only underscores since we have some variable names with underscore. Would it make sense to go the other direction and only use dot notation?

It this referring to the underscore after the `APIGEE` prefix?

rockspore:
Sounds good.

theganyo (approved):

theganyo (approved):

PR #200, opened 2021-04-15
Fix up new Environment configuration schema

josheinhorn:
API Product matching requires an `api_source`. This is currently done using the hostname of the incoming request, however the new Proxy configuration is already associated with an API Proxy, which is semantically the same as `api_source`. This change adds a field to `ProxyConfig` to support this use case.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -240,14 +240,14 @@ type EnvironmentConfig struct {
 	// Unique ID of the environment config
 	ID string `yaml:"id" json:"id"`
 
-	// A list of API configs
-	API []APIConfig `yaml:"proxies" json:"proxies"`
+	// A list of API configs.
+	APIs []APIConfig `yaml:"apis" json:"apis"`
 }
 
 // APIConfig contains settings for an individual API.
 type APIConfig struct {
 	// ID of the API, used to match the api_source of API Product Operations.
-	ID string `yaml:"name" json:"name"`
+	ID string `yaml:"id" json:"id"`
 	
 	// Base path for this API.
 	BasePath string `yaml:"base_path,omitempty" json:"base_path,omitempty"`
--- a/config/config.go
+++ b/config/config.go
@@ -240,40 +240,40 @@ type EnvironmentConfig struct {
 	// Unique ID of the environment config
 	ID string `yaml:"id" json:"id"`
 
-	// A list of proxy configs
-	Proxies []ProxyConfig `yaml:"proxies" json:"proxies"`
+	// A list of API configs
+	API []APIConfig `yaml:"proxies" json:"proxies"`
 }
 
-// ProxyConfig has the proxy configuration.
-type ProxyConfig struct {
-	// Name of the Proxy, used to match the api_source of API Product Operations.
-	Name string `yaml:"name" json:"name"`
+// APIConfig contains settings for an individual API.
+type APIConfig struct {
+	// ID of the API, used to match the api_source of API Product Operations.
+	ID string `yaml:"name" json:"name"`
 	
-	// Base path for this Proxy.
+	// Base path for this API.
 	BasePath string `yaml:"base_path,omitempty" json:"base_path,omitempty"`
 
-	// The default authentication requirements for this Proxy.
+	// The default authentication requirements for this API.
 	Authentication AuthenticationRequirement `yaml:"authentication,omitempty" json:"authentication,omitempty"`
 
-	// The default consumer authorization requirements for this Proxy.
+	// The default consumer authorization requirements for this API.
 	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
 
 	// The default value for the `x-apigee-target` header that will be appended to all allowed requests. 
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -246,6 +246,9 @@ type EnvironmentConfig struct {
 
 // ProxyConfig has the proxy configuration.
 type ProxyConfig struct {
+	// Name of the Proxy, used to match the api_source of API Product Operations.
+	Name string `yaml:"name,omitempty" json:"name,omitempty"`
```
rockspore, line 250: Should we remove the `omitempty` and make it required here?

rockspore (dismissed):

config/config.go
```diff
@@ -148,7 +148,7 @@ type Config struct {
 	Analytics AnalyticsConfig `yaml:"analytics,omitempty" json:"analytics,omitempty" mapstructure:"analytics,omitempty"`
 	// If EnvConfigs is specified, APIKeyHeader, APIKeyClaim, JWTProviderKey in AuthConfig will be ineffectual.
 	Auth       AuthConfig `yaml:"auth,omitempty" json:"auth,omitempty" mapstructure:"auth,omitempty"`
-	EnvConfigs EnvConfigs `yaml:"env_configs,omitempty" json:"env_configs,omitempty" mapstructure:"env_configs,omitempty"`
+	Environments EnvironmentConfigs `yaml:"environments,omitempty" json:"environments,omitempty" mapstructure:"environments,omitempty"`
```
theganyo, line 151: The term "Environment" is so overloaded at this point - and this isn't even related to the existing one. Can we come up with something that adds clarity? APIGroup? ProxySet? OperationVersions? Some combination of those words, or...?

config/config.go
```diff
@@ -240,45 +240,47 @@ type EnvironmentConfig struct {
...
+	// The default consumer authorization requirements for this API.
 	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
 
-	// Name of the target server for this proxy.
-	Target string `yaml:"target" json:"target"`
+	// The default value for the `x-apigee-target` header that will be appended to all allowed requests. 
+	TargetID string `yaml:"target" json:"target"`
 
-	// A list of Operations, names of which must be unique within the proxy config.
+	// A list of API Operations, names of which must be unique within the API.
```
rockspore, line 264: I missed this in my PR. Do we allow an API to have no operations? What would happen if so?

config/config.go
```diff
@@ -240,45 +240,47 @@ type EnvironmentConfig struct {
...
+	// HTTP matching rules for this Operation. If omitted, this API Operation will match all HTTP requests not matched by another API Operation.
 	HTTPMatches []HTTPMatch `yaml:"http_match,omitempty" json:"http_match,omitempty"`
 
-	// Name of the target server for this operation.
-	Target string `yaml:"target" json:"target"`
+	// The value for the `x-apigee-target` header for this Operation that will be appended to all allowed requests. 
+	TargetID string `yaml:"target,omitempty" json:"target,omitempty"`
 }
 
 // AuthenticationRequirement is the interface defining the authentication requirement.
```
rockspore, line 286: JFYI, I did this wrong here and another wrapper is needed. I fixed this in the ongoing work to load the config. So no action required here.

config/config.go
josheinhorn: `ApigeeEnvironments`? That's what this represents after all. Are the other references to `Environment` not also an Apigee Environment?

config/config.go
josheinhorn: We need at least one, kinda like having at least one API Proxy per Environment.

rockspore: Makes sense. So we need to remove the `omitempty` tag.

config/config.go
theganyo: But this isn't, really, is it? It's more like a Environment Revision or Version. I'm just trying to make this make sense for users and this doesn't have any reference to the Apigee Environment we're operating in (ie. the runtime we're making requests to). So, we have a `Config.Global.Environment` and accept `apigee_environment` from Envoy as metadata to override that. Now we're adding another Environment to the config... it's confusing.

rockspore (dismissed):

config/config.go
josheinhorn: I think this comes down to domain entity mapping. The domain is obviously Apigee, thus I am very reluctant to add a brand new entity (e.g. "API Group"). 

FWIW I could not find any references to an Apigee Environment in config.go, the closest is `TenantConfig.EnvName`, is that what you mean?

josheinhorn: Ah I see at least part of the confusion.... this is probably N configs for the _same_ Environment, thus the "revisions" concept. So what about just `EnvironmentConfigs`?

theganyo: Oops... sorry, yes. I meant `TenantConfig.EnvName`.

josheinhorn: Changed to `EnvironmentConfigs` and updating comments to explain the semantics.

theganyo (approved):

PR #202, opened 2021-04-19
Replace Target with generic request header transformation

josheinhorn:
Target is just a static header to be appended to requests via HTTP headers. While useful, it makes more sense to support this via generic HTTP header modification.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -255,8 +255,8 @@ type ProxyConfig struct {
 	// ConsumerAuthorization defines the proxy-level consumer authorization
 	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
 
-	// Transformation rules applied to HTTP request headers.
-	RequestHeaderTransforms RequestHeaderTransformations `yaml:"request_header_transforms,omitempty" json:"request_header_transforms,omitempty"`
+	// Transformation rules applied to HTTP requests.
+	HTTPRequestTransforms HTTPRequestTransformations `yaml:"http_request_transforms,omitempty" json:"http_request_transforms,omitempty"`
 
 	// A list of Operations, names of which must be unique within the proxy config.
 	Operations []APIOperation `yaml:"operations,omitempty" json:"operations,omitempty"`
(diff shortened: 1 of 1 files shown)


rockspore:
I think this is a very good idea. It give customers flexibility to configure custom routes (or other filters such as RBAC) in Envoy via this.

Should we give more capability to this transformation such that path rewrites can also be done here?

config/config.go
```diff
@@ -255,8 +255,8 @@ type ProxyConfig struct {
 	// ConsumerAuthorization defines the proxy-level consumer authorization
 	ConsumerAuthorization ConsumerAuthorization `yaml:"consumer_authorization,omitempty" json:"consumer_authorization,omitempty"`
 
-	// Name of the target server for this proxy.
-	Target string `yaml:"target" json:"target"`
+	// Transformation rules applied to HTTP request headers.
+	RequestHeaderTransforms RequestHeaderTransformations `yaml:"request_header_transforms,omitempty" json:"request_header_transforms,omitempty"`
```
theganyo, line 259: I like this change a lot. A couple questions: 1. Should this be (or also be) in Operation? 2. The next thing would be a RequestBodyTransformation, so I'm wondering if we have a "RequestTransformations" section or if we leave them top-level?

rockspore: Is ext_authz able to tamper with request body?

josheinhorn: No ext_authz cannot change the request body, but ext_proc can. I don't have a preferences between having multiple top-level transform sections versus grouping them under a single `RequestTransforms` section. In general I prefer whatever keeps the config the flattest, and I think both are 2 levels deep.

josheinhorn: And yes, it is already in the Operation (see line 281).

theganyo: Ok. Grouping or not is fine with since I assume there will only ever be two anyway: Headers and Body.

josheinhorn: We probably also want to do first-class transforms for special values like URL and HTTP method. That makes me think we _should_ add `HTTPRequestTransforms` as a grouping, with everything within. Lemme update.

josheinhorn: PTAL

rockspore (approved):
I will remove all the json tags or replace them with mapstructure ones in m own PR then.

config/config.go
theganyo: Looks good. So we're supporting all these transforms for next release?

theganyo (approved):

theganyo:
> I will remove all the json tags or replace them with mapstructure ones in m own PR then.

@rockspore You could otherwise add that change to this PR.

rockspore:
> > I will remove all the json tags or replace them with mapstructure ones in m own PR then.
> 
> @rockspore You could otherwise add that change to this PR.

I already refactored the new config and new tests into separate files in my ongoing work...I see a chunk of merge conflicts already, so it wouldn't save anything to do it here.

theganyo:
@josheinhorn FYI: We have the PR requestor do the "Squash and merge" and delete the PR branch once approved if they have merge rights. You have the green light. :)

config/config.go
josheinhorn: If time allows!

PR #204, opened 2021-04-20
Unmarshal/marshal, load and validate new configs

### LINKED ISSUE #201: load envconfigs (proxy configs) files via command line flag
See #189 for reference.

Will use the flag `--proxy-config` instead of `--operation-config`

### LINKED ISSUE #203: rename config structs
In refactoring configuration structs into the `config` package, the "Config" suffix on the struct names has become redundant and should be removed.

Special case: rename `APIConfig` to `APIProxy` instead of just `API`.

rockspore:
Close #201 

Close #203


theganyo (dismissed):

config/testdata/good_env_config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    http_match:
+    - path_template: /petstore
+      method: GET
+  - name: op-2
+    http_match:
+    - path_template: /bookshop
+      method: POST
+  http_request_transforms:
+    set_headers:
+      x-apigee-target: target
```
josheinhorn, line 27: JFYI we'll probably use `x-apigee-route` ... not super important for test data, just wanted to share since it's relevant

config/env_config.go
```diff
@@ -0,0 +1,431 @@
...
+	"github.com/apigee/apigee-remote-service-golib/v2/errorset"
+	"gopkg.in/yaml.v3"
+)
+
+// validateEnvConfigs checks if there are
+//   * environment configs with the same ID,
+//   * API configs under the same environment config with the same ID,
+//   * JWT authentication requirement under the same API or operation with the same name
+// and report them as errors
+func (c *Config) validateEnvConfigs() error {
```
josheinhorn, line 33: JFYI we'll need very similar validation logic in the control plane, so let's think about how we can scale this sort of development (don't want to write the code twice).

config/env_config.go
```diff
@@ -0,0 +1,431 @@
...
+	Issuer               string                  `yaml:"issuer" mapstructure:"issuer"`
+	RemoteJWKS           *RemoteJWKS             `yaml:"remote_jwks,omitempty" mapstructure:"remote_jwks,omitempty"`
+	Audiences            []string                `yaml:"audiences,omitempty" mapstructure:"audiences,omitempty"`
+	ForwardPayloadHeader string                  `yaml:"forward_payload_header,omitempty" mapstructure:"forward_payload_header,omitempty"`
+	In                   []APIOperationParameter `yaml:"in" mapstructure:"in"`
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface
+func (j *JWTAuthentication) UnmarshalYAML(node *yaml.Node) error {
+	type Unmarsh JWTAuthentication
```
josheinhorn, line 253: Is this supposed to be `var Unmarsh JWTAuthentication`?

josheinhorn, line 253: Why are we declaring this type alias?

josheinhorn: Oh let me see if I got this right... the alias does not declare the `UnmarshalYAML` func, so it can be used with the default YAML impl, right?

rockspore: We need alias to unmarshal those well-defined fields. Without this, `yaml.Unmarshal` would use this `UnmarshalYAML` method which will cause stack overflow.

rockspore: > Oh let me see if I got this right... the alias does not declare the `UnmarshalYAML` func, so it can be used with the default YAML impl, right?

Precisely. The implementation went thru a number of iterations. When implementing the `MarshalYAML` method, I added all the needed fields into the internal wrapper structs. So we could also just read those fields from the unmarshaled wrapper struct.

config/env_config.go
rockspore: If the validation operates on those two schemas, I am not sure how we can avoid repeating at least some of it. Alternatively and not sure if desired, we could generate this and then validate it in the control plane. I would just need to modify the signature of this method (if needed) and export it.

config/env_config.go
```diff
@@ -0,0 +1,431 @@
...
+	// Header values to set. If a specified header is already present, the value here will overwrite it.
+	SetHeaders map[string]string `yaml:"set_headers,omitempty" mapstructure:"set_headers,omitempty"`
+
+	// Headers to remove. Supports single wildcard globbing e.g. `x-apigee-*`.
+	RemoveHeaders []string `yaml:"remove_headers,omitempty" mapstructure:"remove_headers,omitempty"`
+}
+
+// AuthenticationRequirement defines the authentication requirement. It can be jwt, any or all.
+type AuthenticationRequirement struct {
+	AuthenticationRequirements AuthenticationRequirements `yaml:"-"`
```
theganyo, line 151: Referencing AuthenticationRequirements from an Operation like this will be:  `operation.Authentication.AuthenticationRequirements`. I think it would read better if we just made the field name `Requirements` so the reference name would be: `operation.Authentication.Requirements`.

rockspore: Yes this makes sense. Will do.

theganyo (approved):

config/env_config.go
josheinhorn: Yea this is really just a reminder to think about how we can structure code for minimal redundancy, so focus more on structural correctness (no duplicate IDs) versus semantic correctness e.g. that string values use valid syntax, have reasonable lengths, etc.

And yes I agree we will want to run this validator on the generated config in the control plane, the same as we will do for Envoy as a last-ditch effort to avoid a bad rollout. However the error messages will not be even close to as useful as errors generated in the original input validation layer since they won't have any user-facing context.

rockspore: Yes the current validation simply checks for all the duplication. We can add more checks as needed as it evolves.

In the latest commit, I have exported this function.

theganyo: Can we create a new issue for this if needed and move this forward?

rockspore: Created #205.

theganyo: (Or to put another way: @josheinhorn, are you -1 on merging this?)

theganyo:
We have a recursive structure for JWTAuthentication:

JWTAuthentication.In -> APIOperationParameter.Match -> ParamMatch.JWTClaim -> JWTClaim.Requirement -> JWTAuthentication...

What is our strategy for resolving this?

(Aside: We should change the name or add comments to make it clear exactly what type of thing JWTClaim.Requirement refers to since it's just a string. It's really referring to a JWTAuthentication.Name.)

rockspore:
> We have a recursive structure for JWTAuthentication:
> 
> JWTAuthentication.In -> APIOperationParameter.Match -> ParamMatch.JWTClaim -> JWTClaim.Requirement -> JWTAuthentication...
> 
> What is our strategy for resolving this?
> 
> (Aside: We should change the name or add comments to make it clear exactly what type of thing JWTClaim.Requirement refers to since it's just a string. It's really referring to a JWTAuthentication.Name.)

Not sure if we need to resolve this. A config file will be finite anyway...

theganyo:
Ok, let's just validate that we don't allow a JWT inside of a JWT claim. :)

theganyo (approved):

theganyo (approved):

josheinhorn (approved):

PR #209, opened 2021-04-22
remove as much dependencies as possible from env_config.go and its test file

### LINKED ISSUE #208: remove golib imports from env_config and env_config_test
The env_configs are expected to be generated and should be as dependency-free as possible.

rockspore:
close #208


theganyo:
Please add prominent comments at the top of both files explaining that these should be kept free of other dependencies.

theganyo (approved):

PR #210, opened 2021-04-28
sketched solution: compiles, untested

theganyo:
opening for initial review and discussion only

### DIFF
--- a/config/env_spec_ext.go
+++ b/config/env_spec_ext.go
@@ -1,0 +1,67 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package config defines the API Runtime Control config and provides
+// the config loading and validation functions.
+package config
+
+import (
+	"strings"
+
+	"github.com/apigee/apigee-remote-service-golib/v2/path"
+)
+
+// NewEnvironmentSpecExt creates an EnvironmentSpecExt
+func NewEnvironmentSpecExt(spec *EnvironmentSpec) *EnvironmentSpecExt {
+	apiJwtRequirements := make(map[string][]*JWTAuthentication)
+	apiPathTree := path.NewTree()
+	opPathTree := path.NewTree()
+	for i := range spec.APIs {
+		api := spec.APIs[i]
+
+		jwtRequirements := api.Authentication.AllJWTRequirements()
+		apiJwtRequirements[api.ID] = jwtRequirements
+
+		// tree: base path -> APISpec
+		split := strings.Split(api.BasePath, "/")
+		split = append(split, "**")
+		apiPathTree.AddChild(split, 0, api)
+
+		// tree: api.ID -> method -> path -> APIOperation
+		for i := range api.Operations {
+			op := api.Operations[i]
+			for _, m := range op.HTTPMatches {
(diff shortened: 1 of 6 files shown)


config/env_spec_ext.go
```diff
@@ -0,0 +1,116 @@
...
+		for _, val := range []AuthenticationRequirement(v) {
+			found = append(found, allJWTRequirements(val)...)
+		}
+	}
+	return found
+}
+
+func isEmpty(auth AuthenticationRequirement) bool {
+	switch a := auth.Requirements.(type) {
+	case JWTAuthentication:
```
rockspore, line 100: probably safer to add a default case here. same for other similar places.

config/env_spec_request.go
```diff
@@ -0,0 +1,179 @@
...
+// Package config defines the API Runtime Control config and provides
+// the config loading and validation functions.
+package config
+
+import (
+	"net/url"
+	"strings"
+
+	"github.com/apigee/apigee-remote-service-golib/v2/auth/jwt"
+	envoy "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
```
rockspore, line 24: nit: i'd prefer the alias `authv3`

config/env_spec_request.go
```diff
@@ -0,0 +1,179 @@
...
+
+type jwtClaims map[string]interface{}
+
+type jwtResult struct {
+	claims jwtClaims
+	err    error
+}
+
+// GetAPI uses the base path to return an APISpec
+func (e *EnvironmentSpecRequest) GetAPISpec() *APISpec {
```
rockspore, line 54: nil check?

config/env_spec_request.go
```diff
@@ -0,0 +1,179 @@
...
+			u.Query().Get(string(m))
+		}
+	case JWTClaim:
+		return e.getClaimValue(m)
+	}
+	return ""
+}
+
+func (e *EnvironmentSpecRequest) getClaimValue(claim JWTClaim) string {
+	r, ok := e.jwtResults[claim.Requirement]
```
rockspore, line 99: nil check to avoid panics.

config/env_spec_ext.go
theganyo: Why? A default case is not required by the language, it would function exactly the same, and I don't see any additional clarity to be gained.

rockspore: True. I think I wasn't reading it carefully.

config/env_spec_request.go
theganyo: I started to add the checks, but it's never valid to call these methods on a nil struct and would only happen because of programmer error, not user imput, thus I believe it's better to not pretend the call is valid.

rockspore: I fully agree. It just feels safer to me to reduce the possibility of triggering a panic ;-)

theganyo: It does... but also, It could hide error conditions in the code and cause improper behavior.

rockspore: You could add error to the return values, in which case error checking would have to be done by the caller...

config/env_spec_request.go
```diff
@@ -0,0 +1,309 @@
...
+			value = value[:indx]
+		}
+	case Query:
+		if e.queryValues == nil {
+			q := strings.SplitN(e.request.Attributes.Request.Http.Path, "?", 2)
+			if len(q) > 1 {
+				vals, _ := url.ParseQuery(q[1])
+				e.queryValues = vals
+			}
+			value = e.queryValues.Get(string(m))
```
rockspore, line 135: I haven't fully examined but just to make sure this is purposefully within the `if {}` block?

theganyo: Yup, it's just a lazy initialization of e.queryValues.

util/util_test.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Package protostruct supports operations on the protocol buffer Struct message.
-package util
+package util_test
```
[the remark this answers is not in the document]
theganyo: I used that to enforce tests at the public level. It's an optional feature of the language to use the _test package and I find it useful sometimes.

PR #215, opened 2021-05-11
add setpath/addprefix

### LINKED ISSUE #214: Add prefix field to support adding path prefix to authorized requests.

rockspore:
close #214


config/env_spec.go
```diff
@@ -147,6 +147,15 @@ type HTTPRequestTransformations struct {
 
 	// Headers to remove. Supports single wildcard globbing e.g. `x-apigee-*`.
 	RemoveHeaders []string `yaml:"remove_headers,omitempty" mapstructure:"remove_headers,omitempty"`
+
+	// SetPath transforms path on authorized requests.
+	SetPath SetPath `yaml:"set_path,omitempty" mapstructure:"set_path,omitempty"`
+}
+
+// SetPath configures how a request path will be transformed.
+type SetPath struct {
```
josheinhorn, line 156: URLPathTransformations

config/env_spec.go
```diff
@@ -147,6 +147,15 @@ type HTTPRequestTransformations struct {
...
+	// AddPrefix is the prefix that will be added to the request path
+	AddPrefix string `yaml:"add_prefix,omitempty" mapstructure:"add_prefix,omitempty"`
```
josheinhorn, line 158: Are adjacent slashes merged? E.g. `AddPrefix: /prefix/` combined with path `/foo/bar`, is the result `/prefix//foo/bar` or `/prefix/foo/bar`?

josheinhorn (changes requested):

config/env_spec.go
rockspore: It should do so. Let me document this.

PR #221, opened 2021-05-11
API base path must be stripped from request path

### LINKED ISSUE #220: API base path must be stripped from request path
Apigee automatically strips the proxy base path prefix from the request before forwarding, we should do the same.

theganyo:
Fixes #220

### DIFF
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -48,6 +48,7 @@ const (
 	envContextKey        = "apigee_environment"
 	apiContextKey        = "apigee_api"
 	envSpecContextKey    = "apigee_env_config"
+	envoyPathHeader      = ":path"
 )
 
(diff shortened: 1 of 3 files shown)


server/authorization.go
```diff
@@ -277,6 +279,10 @@ func addHeaderTransforms(req *envoy_auth.CheckRequest, envRequest *config.Enviro
 				}
 			}
 
+			// strip proxy base path from request path
+			addHeaderValueOption(okResponse, envoyPathHeader, envRequest.GetOperationPath(), false)
```
rockspore, line 283: This is currently done as the first so a potential `:path` override can happen subsequently. Just to make sure, do we allow this and in general do we allow customers to configure pseudo-header transformations?

PR #236, opened 2021-05-18
append headers support both map and slice of key/values

### LINKED ISSUE #233: AppendHeaders must support multiple values
As header keys can be repeated multiple values, we need to change HttpRequestTransformations.AppendHeaders to represent an array of values. For easy readability, we can leave the existing `key: value` structure in place and add a oneof that also allows `key: []value`. For example, both of these would be allowed in the YAML:

```
AppendHeaders:
  myheader: myvalue
```
and
```
AppendHeaders:
  myheader:
  - value 1
  - value 2
```

rockspore:
close #233


rockspore:
The change is minimal from user's perspective, when they write yamls.

theganyo (approved):

PR #238, opened 2021-05-18
add additional debug logging

### LINKED ISSUE #224: add logging for env spec handling
Audit logging and any debug and/or info level logging as necessary for supportability.

theganyo:
Closes #224

### DIFF
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -24,8 +24,11 @@ import (
 	"github.com/apigee/apigee-remote-service-golib/v2/auth"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth/jwt"
 	"github.com/apigee/apigee-remote-service-golib/v2/log"
+	"github.com/apigee/apigee-remote-service-golib/v2/util"
 	authv3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
 )
+
+const TruncateDebugRequestValuesAt = 5
 
 // NewEnvironmentSpecRequest creates a new EnvironmentSpecRequest
 func NewEnvironmentSpecRequest(authMan auth.Manager, e *EnvironmentSpecExt, req *authv3.CheckRequest) *EnvironmentSpecRequest {
@@ -129,7 +132,7 @@ func (e *EnvironmentSpecRequest) GetParamValue(param APIOperationParameter) stri
 		if indx := strings.Index(value, ","); indx > 0 {
 			value = value[:indx]
 		}
-		log.Debugf("param from header %q: %q", key, value)
+		log.Debugf("param from header %q: %q", key, util.Truncate(value, TruncateDebugRequestValuesAt))
 	case Query:
 		if e.queryValues == nil {
 			q := strings.SplitN(e.request.Attributes.Request.Http.Path, "?", 2)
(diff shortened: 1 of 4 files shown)


config/env_spec_request.go
```diff
@@ -122,23 +122,28 @@ func (e *EnvironmentSpecRequest) GetParamValue(param APIOperationParameter) stri
...
 	case Header:
-		value = e.request.Attributes.Request.Http.Headers[strings.ToLower(string(m))]
+		key := strings.ToLower(string(m))
+		value = e.request.Attributes.Request.Http.Headers[key]
 		// Per Envoy: If multiple headers share the same key, they are merged per HTTP spec.
 		// So, we're just grabbing the first value (up to any comma).
 		if indx := strings.Index(value, ","); indx > 0 {
 			value = value[:indx]
 		}
+		log.Debugf("param from header %q: %q", key, value)
```
rockspore, line 132: Is it okay to print the key as is? Should we redact it?

theganyo: Yeah, I wondered about that myself. I sort of convinced myself that logging it only at debug level might be okay, but we really have no idea which parameters are sensitive. So while it will make debugging more difficult, perhaps we should not log any of these params - even at debug level.

theganyo: Otherwise, we could potentially severely truncate all the values to just a few characters.

theganyo: Also... this applies to header manipulation logging as well.

server/authorization.go
```diff
@@ -305,7 +306,32 @@ func addHeaderTransforms(req *authv3.CheckRequest, envRequest *config.Environmen
...
+	var b strings.Builder
+	b.WriteString("Request header mods:\n")
+	if len(okResponse.Headers) > 0 {
+		for _, h := range okResponse.Headers {
+			addAppend := "="
+			if h.Append.Value {
+				addAppend = "+"
+			}
+			b.WriteString(fmt.Sprintf("  %s %q: %q\n", addAppend, h.Header.Key,
+				golibutil.Truncate(h.Header.Value, config.TruncateDebugRequestValuesAt)))
```
rockspore, line 325: Not specific to this PR but I wonder if we should merge two util packages in -golib. WDYT?

rockspore (approved):

config/env_spec_request.go
rockspore: Yeah truncating values sounds good.

server/authorization.go
theganyo: It's a little awkward, but it doesn't make sense to add envoy / proto imports to -golib.

rockspore: Technically, the util here only imports the generic protobuf types and has nothing specific to Envoy APIs. It's fine to me as is though.

theganyo: Ok, let's just leave it for now.

server/authorization_test.go
```diff
@@ -288,6 +296,59 @@ func TestEnvRequestCheck(t *testing.T) {
 	}
 }
 
+// func TestLogHeaderValueOptions(t *testing.T) {
```
rockspore, line 299: Removing this if not needed?

PR #241, opened 2021-05-20
read a list of files or directories for env spec

### LINKED ISSUE #240: Read env spec from a folder instead of a list of files via commang line flag

rockspore:
close #240


rockspore:
> Would it make sense to allow folders refs directly in c.EnvironmentSpecs.References instead of in main?

I would like that. It's just that would change the `API`. If that's fine to everyone, I can change it.

theganyo:
Well, the API is just a string containing a path, so it's just the comment. :)

theganyo (approved):

PR #265, opened 2021-06-17
validate method and match any

### LINKED ISSUE #263: Validate HTTPMatch.Method method against valid list
Valid list from API Product Operations: GET, PATCH, POST, PUT, DELETE

(Note: OPTIONS will handled as a special case in planned CORS support.)

theganyo:
Fixes #263

### DIFF
--- a/config/env_spec.go
+++ b/config/env_spec.go
@@ -25,6 +25,12 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const anyMethod = ""
+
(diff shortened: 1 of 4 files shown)


rockspore (approved):
This LGTM. But if we decided to define "all" methods to exclude special ones like "HEAD", "OPTIONS", etc, this wildcard approach would seem to break.

theganyo:
Yes, but doing so would require code changes in any case.

PR #272, opened 2021-06-23
If there are no operations for an api, assume a "match all" operation

### LINKED ISSUE #271: If there are no operations for an api, assume a "match all" operation
Currently, an api without operations will always fail as there is no match. Instead, it should succeed as if an operation exists that allows all methods and paths.

theganyo:
Fixes #271

### DIFF
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -30,6 +30,11 @@ import (
 
 const TruncateDebugRequestValuesAt = 5
 
+// a "match all" operation for apis without operations
+var synthesizedOperation = &APIOperation{
+	Name: "*",
+}
+
 // NewEnvironmentSpecRequest creates a new EnvironmentSpecRequest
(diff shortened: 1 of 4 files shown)


config/env_spec_request.go
```diff
@@ -30,6 +30,11 @@ import (
 
 const TruncateDebugRequestValuesAt = 5
 
+// a "match all" operation for apis without operations
+var synthesizedOperation = &APIOperation{
+	Name: "*",
```
rockspore, line 35: Just wanted to confirm, do we indeed disallow such a name?

rockspore: nvm, I realize it doesn't matter.

rockspore (approved):

config/env_spec_request.go
theganyo: We don't validate against it, but since it's only used when there are no operations, it can't conflict with anything.

PR #277, opened 2021-06-29
strip base path off for proxy mode

### LINKED ISSUE #276: In `proxy` mode, API products' operations' path should be matched against subpath
subpath refers to the path after the basepath is stripped.

rockspore:
close #276

### DIFF
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -131,6 +131,11 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 		}
 		log.Debugf("operation: %s", operation.Name)
 
+		// strip the basepath off path
+		if a.handler.operationConfigType == product.ProxyOperationConfigType {
+			path = envRequest.GetOperationPath()
+		}
+
 		if !envRequest.IsAuthenticated() {
 			log.Debugf("authentication requirements not met")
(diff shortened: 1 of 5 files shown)


server/authorization.go
```diff
@@ -131,6 +131,11 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 		}
 		log.Debugf("operation: %s", operation.Name)
 
+		// strip the basepath off path
```
theganyo, line 134: Nit: This could be moved to after the envRequest.IsAuthenticated() check below as it's not needed if that check fails.

server/authorization_test.go
```diff
@@ -344,6 +344,75 @@ func TestEnvRequestCheck(t *testing.T) {
 	}
 }
 
+func TestBasePathStripping(t *testing.T) {
```
theganyo, line 347: We should probably test the converse as well - that the path should not be stripped for non-proxy type.

server/authorization.go
rockspore: Sure.

server/authorization_test.go
rockspore: Added tests

theganyo (approved):

PR #293, opened 2021-07-16
return 503 from ext_auth until products loaded

### LINKED ISSUE #292: do not open ext_auth request listener until ready

theganyo:
Fixes #292

### DIFF
--- a/server/accesslog_test.go
+++ b/server/accesslog_test.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/apigee/apigee-remote-service-golib/v2/analytics"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth"
-	"github.com/apigee/apigee-remote-service-golib/v2/util"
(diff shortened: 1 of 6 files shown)


server/accesslog_test.go
```diff
@@ -120,6 +121,7 @@ func TestHandleHTTPAccessLogs(t *testing.T) {
 			orgName:      extAuthzFields[headerOrganization].GetStringValue(),
 			envName:      extAuthzFields[headerEnvironment].GetStringValue(),
 			analyticsMan: testAnalyticsMan,
+			ready:        util.NewAtomicBool(true),
```
rockspore, line 124: Why do we need this field here if the access log is always listening? Same Q for the other line in this file.

theganyo: Yeah, we don't. Just wasn't paying close attention to it. :)

rockspore (approved):

PR #296, opened 2021-07-23
gcp managed internal api will not be exclusive to AX credentials

### LINKED ISSUE #295: Accepts gcp managed internal API
`Config.Tenant.InternalAPI` with suffix `googleapis.com` should not be considered as exclusive to analytics credentials.

rockspore:
close #295

### DIFF
--- a/server/handler.go
+++ b/server/handler.go
@@ -212,10 +212,7 @@ func NewHandler(cfg *config.Config) (*Handler, error) {
 		analyticsClient = clientAuthorizedByCredentials(cfg, "analytics", cfg.Analytics.Credentials)
 		// overwrite the internalAPI to the GCP managed host if not initialized yet
 		if internalAPI == nil {
-			internalAPI, err = url.Parse(config.GCPExperienceBase)
-			if err != nil {
-				return nil, err
-			}
+			internalAPI, _ = url.Parse(config.GCPExperienceBase)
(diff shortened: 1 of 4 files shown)


server/handler.go
```diff
@@ -210,8 +210,13 @@ func NewHandler(cfg *config.Config) (*Handler, error) {
 	if cfg.Analytics.Credentials != nil {
 		// Attempts to get an authorized http client with given analytics credentials
 		analyticsClient = clientAuthorizedByCredentials(cfg, "analytics", cfg.Analytics.Credentials)
-		// overwrite the internalAPI to the GCP managed host
-		internalAPI, _ = url.Parse(config.GCPExperienceBase)
+		// overwrite the internalAPI to the GCP managed host if not initialized yet
+		if internalAPI == nil {
+			internalAPI, err = url.Parse(config.GCPExperienceBase)
+			if err != nil {
```
theganyo, line 216: pointless check for a constant

theganyo (dismissed):

server/handler.go
rockspore: Yeah. Let me remove it again. Will need new approval then :)

theganyo: ok, sure :)

theganyo (approved):

PR #312, opened 2021-08-13
always strip basepath for envspec configs

### LINKED ISSUE #276: In `proxy` mode, API products' operations' path should be matched against subpath

theganyo:
close #276

### DIFF
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -152,9 +152,7 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 		}
 
 		// strip the basepath off path
-		// if a.handler.operationConfigType == product.ProxyOperationConfigType {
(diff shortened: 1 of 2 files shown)


server/authorization.go
```diff
@@ -152,9 +152,9 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 		}
 
 		// strip the basepath off path
-		if a.handler.operationConfigType == product.ProxyOperationConfigType {
-			path = strings.SplitN(envRequest.GetOperationPath(), "?", 2)[0]
-		}
+		// if a.handler.operationConfigType == product.ProxyOperationConfigType {
```
rockspore, line 155: Why not strip these commented code off?

rockspore (approved):

server/authorization.go
theganyo: Forgetfulness. :)

PR #316, opened 2021-08-23
Implement http rewrites

### LINKED ISSUE #227: Implement http rewrites

theganyo:
Closes #227

### DIFF
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -430,7 +430,7 @@ func (e *EnvironmentSpecRequest) IsAuthorizationRequired() bool {
 	return !e.GetConsumerAuthorization().Disabled && !e.GetConsumerAuthorization().isEmpty()
 }
 
-func (e *EnvironmentSpecRequest) GetHTTPRequestTransformations() (transforms HTTPRequestTransforms) {
+func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequestTransforms) {
 	if e != nil {
 		op := e.GetOperation()
 		if op != nil && !op.HTTPRequestTransforms.isEmpty() {
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/apigee/apigee-remote-service-envoy/v2/transform"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth/jwt"
 	"github.com/apigee/apigee-remote-service-golib/v2/log"
@@ -46,6 +47,14 @@ const (
 	CORSMaxAge                = "access-control-max-age"
 	CORSAllowCredentials      = "access-control-allow-credentials"
 	CORSAllowCredentialsValue = "true"
+
+	VariableNamespaceSeparator = "."
+	RequestNamespace           = "request"
+	QueryNamespace             = "query"
+	PathNamespace              = "path"
+	HeaderNamespace            = "headers"
+	RequestPath                = "path"
+	RequestQuerystring         = "querystring"
 )
 
 // a "match all" operation for apis without operations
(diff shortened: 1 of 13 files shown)


config/env_spec_request.go
```diff
@@ -71,12 +81,57 @@ type EnvironmentSpecRequest struct {
...
 	operation             *APIOperation
-	queryValues           url.Values
-	operationPath         string
 	consumerAuthorization *ConsumerAuthorization
+	variables             *requestVariables // for template reification
+}
+
+func (e *EnvironmentSpecRequest) parseRequest() {
+
+	path, queryString := func() (string, string) {
```
rockspore, line 92: What is the benefit of having an anonymous func here?

config/env_spec_request.go
```diff
@@ -86,6 +141,88 @@ type jwtResult struct {
...
+
+	vars := &requestVariables{
+		path:    pathTemplate.Extract(opPath),
+		headers: e.Request.Attributes.Request.Http.Headers,
+		request: map[string]string{},
+		query:   map[string]string{},
+	}
+
+	vars.request[RequestPath] = opPath
+	vars.request[RequestQuerystring] = queryString
```
rockspore, line 154: Why do we need to store this?

config/env_spec_request.go
```diff
@@ -324,7 +430,7 @@ func (e *EnvironmentSpecRequest) IsAuthorizationRequired() bool {
 	return !e.GetConsumerAuthorization().Disabled && !e.GetConsumerAuthorization().isEmpty()
 }
 
-func (e *EnvironmentSpecRequest) GetHTTPRequestTransformations() (transforms HTTPRequestTransformations) {
+func (e *EnvironmentSpecRequest) GetHTTPRequestTransformations() (transforms HTTPRequestTransforms) {
```
rockspore, line 433: nit: should we change the function name to `...Transforms`?

server/authorization.go
```diff
@@ -354,24 +353,57 @@ func addRequestHeaderTransforms(req *authv3.CheckRequest, envRequest *config.Env
...
+			pathTransform := transforms.PathTransform
+			var targetPath = envRequest.GetOperationPath()
+			if pathTransform != "" {
+				targetPath = envRequest.Reify(pathTransform)
+				targetPath = path.Clean(targetPath)
+			}
+
+			queryMap := envRequest.GetQueryParams()
+			for _, t := range transforms.QueryTransforms.Remove {
+				t = strings.ToLower(t) // TODO: move
```
rockspore, line 368: What does this TODO mean?

server/authorization.go
```diff
@@ -354,24 +353,57 @@ func addRequestHeaderTransforms(req *authv3.CheckRequest, envRequest *config.Env
...
+					}
+				}
+				targetPath = targetPath + "?" + strings.Join(queryParams, "&")
+			}
+
+			addRequestHeader(okResponse, envoyPathHeader, targetPath, false)
+
+			// header transforms
+			for _, t := range transforms.HeaderTransforms.Remove {
+				t = strings.ToLower(t) // TODO: move
```
rockspore, line 397: same here

config/env_spec_request.go
theganyo: Just encapsulation & self-documentation... avoiding some variable pollution. Do you think it would be clearer to remove it?

rockspore: No it's good as is ;-)

config/env_spec_request.go
theganyo: The "request.path" variable is just the path part, the "request.querystring" is just the query string part. The idea is to give users options to deal with the larger parts of the path instead of just the individually parsed pieces. For example, I might not want to be able to specify the entire request query string when I design the API... so this lets me just use the whole thing.

config/env_spec_request.go
theganyo: done

config/env_spec_request.go
rockspore: I see. Thanks.

server/authorization.go
theganyo: Good question. I should've removed the TODO, but my thought was to move the ToLower() to parsing, but that would add more complexity without much value.

rockspore (approved):

PR #324, opened 2021-09-10
Append "/" if `opPath` has no "/" as prefix.

### LINKED ISSUE #323: If basepath is specified as /, then it should not be stripped.
This will result in an unexpected HTTP path

rockspore:
fix #323


theganyo:
Do we need a test on this?

rockspore:
> Do we need a test on this?

Absolutely. I was looking for a way to add variational change to minimize work and now I found one! Without the change, the `TestVariables` test will now break.

theganyo (approved):

PR #329, opened 2021-09-15
change gatewaysource in AX record

### LINKED ISSUE #328: "gateway_source" in AX records should be "Configurable" for EG4A

rockspore:
fix #328

### DIFF
--- a/server/accesslog.go
+++ b/server/accesslog.go
@@ -23,6 +23,7 @@ import (
 	"github.com/apigee/apigee-remote-service-golib/v2/analytics"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth"
 	"github.com/apigee/apigee-remote-service-golib/v2/log"
+	"github.com/apigee/apigee-remote-service-golib/v2/product"
 	als "github.com/envoyproxy/go-control-plane/envoy/service/accesslog/v3"
 	"github.com/golang/protobuf/ptypes/duration"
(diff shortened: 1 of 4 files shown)


server/accesslog.go
```diff
@@ -50,6 +51,10 @@ func (a *AccessLogServer) Register(s *grpc.Server, handler *Handler, d time.Dura
 	a.handler = handler
 	a.streamTimeout = d
 	a.context = ctx
+	a.gatewaySource = defaultGatewaySource
+	if a.handler.operationConfigType == "PROXY" {
```
theganyo, line 55: I believe this should use the `product.ProxyOperationConfigType` constant, should it not?

server/accesslog_test.go
```diff
@@ -121,6 +121,7 @@ func TestHandleHTTPAccessLogs(t *testing.T) {
 			envName:      extAuthzFields[headerEnvironment].GetStringValue(),
 			analyticsMan: testAnalyticsMan,
 		},
+		gatewaySource: "configurable",
```
theganyo, line 124: Use a constant for all these instances of string "configurable"?

server/accesslog.go
rockspore: Done.

server/accesslog_test.go
rockspore: Ack.

theganyo (approved):

PR #335, opened 2021-10-05
add google oauth

### LINKED ISSUE #334: Support Google OAuth in environment spec
Add the relevant struct and fields to the config.

rockspore:
close #334

### DIFF
--- a/config/env_spec.go
+++ b/config/env_spec.go
@@ -205,6 +205,13 @@ type APISpec struct {
 	// CORS Policy
 	Cors CorsPolicy `yaml:"cors,omitempty" mapstructure:"cors,omitempty"`
 
+	// Service account that will be impersonated for backend access.
+	ServiceAccountEmail string `yaml:"service_account_email,omitempty" mapstructure:"service_account_email,omitempty"`
+
+	// GoogleOAuth configures how to authenticate the request to the backend.
+	// Can be overridden at the operation level.
+	GoogleOAuth GoogleOAuth `yaml:"google_oauth,omitempty" mapstructure:"google_oauth,omitempty"`
+
 	// JWTAuthentication.Name -> *JWTAuthentication
 	jwtAuthentications map[string]*JWTAuthentication `yaml:"-" mapstructure:"-"`
 }
@@ -225,6 +232,10 @@ type APIOperation struct {
 
 	// Transformation rules applied to HTTP requests for this Operation. Overrides the rules set at the API level.
 	HTTPRequestTransforms HTTPRequestTransforms `yaml:"http_request_transforms,omitempty" mapstructure:"http_request_transforms,omitempty"`
+
+	// GoogleOAuth configures how to authenticate the request to the backend.
+	// Overrides the API (proxy) level settings.
+	GoogleOAuth GoogleOAuth `yaml:"google_oauth,omitempty" mapstructure:"google_oauth,omitempty"`
 
 	// JWTAuthentication.Name -> *JWTAuthentication
 	jwtAuthentications map[string]*JWTAuthentication `yaml:"-" mapstructure:"-"`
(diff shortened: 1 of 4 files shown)


config/env_spec.go
```diff
@@ -602,3 +611,101 @@ type CorsPolicy struct {
...
+type TargetAuthentication struct {
+	// The time between two adjacent token refreshments.
+	RefreshInterval time.Duration `yaml:"refresh_interval,omitempty" mapstructure:"refresh_interval,omitempty"`
+
+	// GoogleOAuth configures how to authenticate the request to the backend with Google OAuth tokens.
+	GoogleOAuth GoogleOAuth `yaml:"google_oauth,omitempty" mapstructure:"google_oauth,omitempty"`
+
+	// TokenSource is used to fetch OAuth token.
+	// This is not part of the yaml config and will be generated internally.
+	TokenSource oauth2.TokenSource `yaml:"-" mapstructure:"-"`
```
theganyo, line 625: What is the purpose of this?

config/env_spec.go
```diff
@@ -602,3 +611,101 @@ type CorsPolicy struct {
...
 	return len(c.AllowOrigins) == 0 && len(c.AllowOriginsRegexes) == 0
 }
+
+// TargetAuthentication configures how to authenticate the request to the backend.
+type TargetAuthentication struct {
+	// The time between two adjacent token refreshments.
+	RefreshInterval time.Duration `yaml:"refresh_interval,omitempty" mapstructure:"refresh_interval,omitempty"`
+
+	// GoogleOAuth configures how to authenticate the request to the backend with Google OAuth tokens.
+	GoogleOAuth GoogleOAuth `yaml:"google_oauth,omitempty" mapstructure:"google_oauth,omitempty"`
```
theganyo, line 621: Should this struct be a oneof, so it could be a GoogleOAuth or another kind? Maybe something like OAuthProvider or something like that?

config/env_spec.go
rockspore: This is to generate the oauth token. I am implementing it in a different pkg. If we have other providers, we can use this same interface. The plan is to initiate this on loading the config so that it is easier to access.

rockspore: That would require maps of maps to track operation level token source. I think this is very similar to the jwtAuthentications, which was put in the config for the relative easiness to track. I can make this internal and add a accessor method if that makes more sense.

theganyo: I see. It seems fine. Without the implementation, it's just hard to see how it's being used.

rockspore: Yeah before merging this, I will show you a draft PR in -golib so it's clearer how token source is configured and consumed.

rockspore:
Moved the oauth token source impls into this PR.

config/env_spec_request_test.go
```diff
@@ -833,6 +836,155 @@ func TestAllowedOrigin(t *testing.T) {
...
+			envoyReq := testutil.NewEnvoyRequest(http.MethodGet, test.path, nil, nil)
+			req := NewEnvironmentSpecRequest(&testAuthMan{}, specExt, envoyReq)
+			got, err := req.TargetAuth()
+			if err != nil {
+				t.Fatalf("TargetAuth() err = %v, wanted no error", err)
+			}
+			if test.wantTargetAuth != got {
+				t.Errorf("TargetAuth() = %q, wanted %q", got, test.wantTargetAuth)
+			}
+			// if g, ok := req.apiSpec.TargetAuthentication.OAuthProvider.(GoogleOAuth); ok {
```
theganyo, line 978: Remove unused code?

server/authorization.go
```diff
@@ -533,7 +569,7 @@ func (a *AuthorizationServer) createConditionalEnvoyDenied(
 
 	if authContext != nil && a.handler.allowUnauthorized {
 		log.Debugf("sending ok (actual: %s)", code.String())
-		return a.createEnvoyForwarded(req, tracker, authContext, api, envRequest)
+		return a.createEnvoyForwarded(req, tracker, authContext, api, envRequest, "TODO")
```
theganyo, line 572: TODO?

oauth/iam/iam.go
```diff
@@ -0,0 +1,231 @@
...
+	ats.mu.Lock()
+	defer ats.mu.Unlock()
+	// Deep copy the object to avoid data race.
+	return &oauth2.Token{
+		AccessToken: ats.token.AccessToken,
+		Expiry:      ats.token.Expiry,
+	}, nil
+}
+
+func (ats *AccessTokenSource) refresh() error {
```
theganyo, line 177: I think you'll want to avoid a potential thundering herd issue in your refresh functions. See: https://github.com/apigee/apigee-remote-service-golib/blob/0bc2b615ebe95fb9b436bcdf765cac2e04ba66c2/auth/key/verify_api_key.go#L174.

Also, may need backoff() function to avoid rapid repeated failed callouts and cancel()  to allow quicker shutdown. See: https://github.com/apigee/apigee-remote-service-golib/blob/0bc2b615ebe95fb9b436bcdf765cac2e04ba66c2/auth/key/verify_api_key.go#L205.

oauth/iam/iam.go
```diff
@@ -0,0 +1,231 @@
...
+	defer its.mu.Unlock()
+	// Deep copy the object to avoid data race.
+	return &oauth2.Token{
+		AccessToken: its.token.AccessToken,
+		Expiry:      its.token.Expiry,
+	}, nil
+}
+
+func (its *IdentityTokenSource) refresh() error {
+	req := &iamv1.GenerateIdTokenRequest{
```
theganyo, line 215: See refresh() comments above.

oauth/iam/iam.go
theganyo: I believe the following is the case, is it not? Each incoming request calls EnvironmentSpecRequest.TargetAuth() which calls Token() which can call refresh() which will call GenerateAccessToken. You could have 1:1 such calls with incoming requests along this tree.

rockspore: Ah that makes sense! Added the herbBuster logic. PTAL.

theganyo (changes requested):
Removing approval based on offline discussion.

PR #338, opened 2021-10-15
Support native gRPC services in environment config

emschwar:
* If you set grpc_service on an API, that API supports a gRPC API with
  that name.
* Authn and authz can be set on the API as normal.
* To set custom authn, authz, or target server for a given method,
  use that method's name as the Operation ID.

  For example, to support custom authorization for `echo.EchoService.ListPets`:

  ```
  id: env_id
  apis:
  - id: grpc-api
    grpc_service: echo.Echoservice
    operations:
    - id: ListPets
      consumer_authorization:
        # ...
  ```
* A gRPC service may be specified in only one API per environment.
* `basepath` is optional for an API if `grpc_service` is specified.


config/env_spec_request.go
```diff
@@ -54,6 +57,7 @@ const (
 	PathNamespace              = "path"
 	HeaderNamespace            = "headers"
 	RequestPath                = "path"
+	OriginalPath               = "original_path"
```
theganyo, line 60: Hmm... "path" and "original_path" seems like it will be confusing because "path" is also the original path without the base path, and there will be a further target path after it's processed. Rethinking the naming here would be worthwhile, I think.

config/env_spec_ext.go
```diff
@@ -96,15 +109,25 @@ func NewEnvironmentSpecExt(spec *EnvironmentSpec) (*EnvironmentSpecExt, error) {
...
+			isGRPC := api.GrpcService != ""
 			op := api.Operations[i]
 
-			if len(op.HTTPMatches) == 0 { // empty is wildcard
-				split = []string{api.ID, wildcard, wildcard}
+			// For gRPC APIs, always interpret the op Name as a gRPC method, so add it as a
+			// child of the API.
+
+			if isGRPC {
+				split := []string{api.ID, "POST", "", op.Name}
```
rockspore, line 119: Why is there an empty string between `"POST"` and `op.Name`?

theganyo: Right, empty strings are skipped in the tree and during matching. But if for some reason it is necessary here, we should add a comment explaining why.

theganyo: Right, empty strings are skipped in the tree and during matching. But if for some reason it is necessary here, we should add a comment explaining why.

theganyo (approved):

PR #341, opened 2021-10-20
Define context variable and google iam credentials

rockspore:


theganyo:
We need to keep documenting the PRs and associating them with issues. So, please add a description. Also, I think we should associate this PR with an omnibus issue that describes the entire scope of work - and perhaps a breakdown of the expected work PRs. It's often extremely useful to have this historical context.

PR #358, opened 2021-11-19
Add RevisionID to APISpec

### LINKED ISSUE #357: Feature request: Add support for revision ID for APISpec
You have an idea? We'd love to hear it!

## What's missing?
When reporting errors related to an APISpec, it would be helpful to know the revision of the API Spec that generated the error, in case the two are correlated.

## Your preferred solution
Add a field in APISpec to reflect its revision ID.

emschwar:
* In order to avoid passing yet more arguments into apigeeDynamicDataHeaders(),
  replace the `basepath` arg with the APISpec as a whole, so that we can pull
  whatever other data we need out of it as we go on.

Closes #357

### DIFF
--- a/config/env_spec.go
+++ b/config/env_spec.go
@@ -226,7 +226,7 @@ type APISpec struct {
 	ID string `yaml:"id" mapstructure:"id"`
 
 	// RevisionID of the API, used to report the revision of the spec that generated errors.
-	RevisionID string `yaml:"revision_id" mapstructure:"revision_id,omitempty"`
+	RevisionID string `yaml:"revision_id" mapstructure:"revision_id"`
 
 	// Name of the gRPC service provided by this API. Used to map native gRPC method calls.
 	GrpcService string `yaml:"grpc_service,omitempty" mapstructure:"grpc_service,omitempty"`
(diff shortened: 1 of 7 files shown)


config/env_spec.go
```diff
@@ -225,6 +225,8 @@ type APISpec struct {
 	// ID of the API, used to match the api_source of API Product Operations.
 	ID string `yaml:"id" mapstructure:"id"`
 
+	RevisionID string `yaml:"revision_id" mapstructure:"revision_id,omitempty"`
```
rockspore, line 228: Please add comments with `// RevisionID ...`

rockspore: Also remove the omitempty in the mapstructure tag.

rockspore:
The test is also failing, please take a look.

config/env_spec.go
emschwar: Done.

emschwar:
> The test is also failing, please take a look.

I got it passing-- maybe not the best way; let me know if it needs a fix.

config/env_spec.go
rockspore: nit: please remove the omitempty part

rockspore (approved):

config/env_spec.go
emschwar: Done.

PR #363, opened 2021-12-08
Define and Use the New AdapterFault struct for envoy adapter

kansalarchit:
Define a new AdapterFault struct for envoy adapter. 
Replace the bool fault with the new AdapterFault to make fault reporting dynamic.

### DIFF
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -372,7 +372,7 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 		if result.err == nil {
 			return nil
 		} else {
-			fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+			return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 		}
 	}
 
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -354,9 +354,9 @@ func (e *EnvironmentSpecRequest) JWTAuthentications() []*JWTAuthentication {
 // looks up the JWTAuthentication by name and runs verification
 // returns nil if found and verified
 // any error can be in e.jwtResults[name]
-func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
+func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) *fault.AdapterFault {
 	if e == nil {
-		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	}
 	var jwtReq *JWTAuthentication
 	if len(e.GetOperation().jwtAuthentications) > 0 {
@@ -366,13 +366,13 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 	}
 	if jwtReq == nil {
 		log.Debugf("JWTAuthentication %q not found", name)
-		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	}
 	if result := e.jwtResults[name]; result != nil { // return from cache
 		if result.err == nil {
 			return nil
 		} else {
-			return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+			return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 		}
 	}
 
@@ -423,7 +423,7 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 		}
 	}
 
-	return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+	return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 }
 
 // returns error if passed value is not in claim as string or []string
@@ -448,7 +448,7 @@ func mustBeInClaim(value, name string, claims map[string]interface{}) error {
 
 // IsAuthenticated returns nil if AuthenticatationRequirements are met for the request
 // Returns nil if AuthenticatationRequirements are empty or disabled.
-func (e *EnvironmentSpecRequest) IsAuthenticated() error {
+func (e *EnvironmentSpecRequest) IsAuthenticated() *fault.AdapterFault {
 	return e.meetsAuthenticatationRequirements(e.getAuthenticationRequirement())
 }
 
@@ -486,9 +486,9 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 }
 
 // returns nil if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) error {
+func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) *fault.AdapterFault {
 	if e == nil {
-		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	}
 	if auth.Requirements == nil || auth.Disabled {
 		return nil
@@ -502,16 +502,16 @@ func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth Authenti
 				return nil
 			}
 		}
-		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	case AllAuthenticationRequirements:
 		for _, r := range []AuthenticationRequirement(a) {
 			if e.meetsAuthenticatationRequirements(r) != nil {
-				return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+				return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 			}
 		}
 		return nil
 	default:
-		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	}
 }
 
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -449,7 +449,7 @@ func mustBeInClaim(value, name string, claims map[string]interface{}) error {
 // Authenticate returns nil if AuthenticatationRequirements are met for the request
 // Returns nil if AuthenticatationRequirements are empty or disabled.
 func (e *EnvironmentSpecRequest) Authenticate() error {
-	return e.meetsAuthenticatationRequirements(e.getAuthenticationRequirement())
+	return e.verifyAuthenticatationRequirements(e.getAuthenticationRequirement())
 }
 
 func (e *EnvironmentSpecRequest) getAuthenticationRequirement() (auth AuthenticationRequirement) {
@@ -486,7 +486,7 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 }
 
 // returns nil if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) error {
+func (e *EnvironmentSpecRequest) verifyAuthenticatationRequirements(auth AuthenticationRequirement) error {
 	if e == nil {
 		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	}
@@ -498,14 +498,14 @@ func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth Authenti
 		return e.verifyJWTAuthentication(a.Name)
 	case AnyAuthenticationRequirements:
 		for _, r := range []AuthenticationRequirement(a) {
-			if e.meetsAuthenticatationRequirements(r) == nil {
+			if e.verifyAuthenticatationRequirements(r) == nil {
 				return nil
 			}
 		}
 		return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 	case AllAuthenticationRequirements:
 		for _, r := range []AuthenticationRequirement(a) {
-			if e.meetsAuthenticatationRequirements(r) != nil {
+			if e.verifyAuthenticatationRequirements(r) != nil {
 				return fault.CreateAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
 			}
 		}
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -22,12 +22,14 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/apigee/apigee-remote-service-envoy/v2/fault"
 	"github.com/apigee/apigee-remote-service-envoy/v2/transform"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth"
 	"github.com/apigee/apigee-remote-service-golib/v2/auth/jwt"
 	"github.com/apigee/apigee-remote-service-golib/v2/log"
 	"github.com/apigee/apigee-remote-service-golib/v2/util"
 	authv3 "github.com/envoyproxy/go-control-plane/envoy/service/auth/v3"
+	"github.com/gogo/googleapis/google/rpc"
 )
 
 const TruncateDebugRequestValuesAt = 5
@@ -350,11 +352,11 @@ func (e *EnvironmentSpecRequest) JWTAuthentications() []*JWTAuthentication {
 }
 
 // looks up the JWTAuthentication by name and runs verification
-// returns true if found and verified
+// returns nil if found and verified
 // any error can be in e.jwtResults[name]
-func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) bool {
+func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 	if e == nil {
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 	}
 	var jwtReq *JWTAuthentication
 	if len(e.GetOperation().jwtAuthentications) > 0 {
@@ -364,10 +366,14 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) bool {
 	}
 	if jwtReq == nil {
 		log.Debugf("JWTAuthentication %q not found", name)
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 	}
 	if result := e.jwtResults[name]; result != nil { // return from cache
-		return result.err == nil
+		if result.err == nil {
+			return nil
+		} else {
+			fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
+		}
 	}
 
 	// uncached, parse it
@@ -413,11 +419,11 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) bool {
 		setResult(claims, err)
 		// First match wins
 		if err == nil {
-			return true
+			return nil
 		}
 	}
 
-	return false
+	return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 }
 
 // returns error if passed value is not in claim as string or []string
@@ -440,9 +446,9 @@ func mustBeInClaim(value, name string, claims map[string]interface{}) error {
 	return fmt.Errorf("%q not in claim %q", value, name)
 }
 
-// IsAuthenticated returns true if AuthenticatationRequirements are met for the request
-// Returns true if AuthenticatationRequirements are empty or disabled.
-func (e *EnvironmentSpecRequest) IsAuthenticated() bool {
+// IsAuthenticated returns nil if AuthenticatationRequirements are met for the request
+// Returns nil if AuthenticatationRequirements are empty or disabled.
+func (e *EnvironmentSpecRequest) IsAuthenticated() error {
 	return e.meetsAuthenticatationRequirements(e.getAuthenticationRequirement())
 }
 
@@ -479,33 +485,33 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
-// returns true if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) bool {
+// returns nil if auth is empty or disabled
+func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) error {
 	if e == nil {
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 	}
 	if auth.Requirements == nil || auth.Disabled {
-		return true
+		return nil
 	}
 	switch a := auth.Requirements.(type) {
 	case JWTAuthentication:
 		return e.verifyJWTAuthentication(a.Name)
 	case AnyAuthenticationRequirements:
 		for _, r := range []AuthenticationRequirement(a) {
-			if e.meetsAuthenticatationRequirements(r) {
-				return true
+			if e.meetsAuthenticatationRequirements(r) == nil {
+				return nil
 			}
 		}
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 	case AllAuthenticationRequirements:
 		for _, r := range []AuthenticationRequirement(a) {
-			if !e.meetsAuthenticatationRequirements(r) {
-				return false
+			if e.meetsAuthenticatationRequirements(r) != nil {
+				return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 			}
 		}
-		return true
+		return nil
 	default:
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
 	}
 }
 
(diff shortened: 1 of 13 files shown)


theganyo:
The copyright should be 2021. But other than that, this struct isn't being used anywhere and has a TODO, so it seems incomplete. What's the context on this?

theganyo:
It's hard to tell if this is PR ready for review or not. Can we move it to "draft" if not, and then back when it's ready?

kansalarchit:
The PR is ready for review. Please take a look.

theganyo (changes requested):

server/header_context.go
```diff
@@ -104,15 +106,21 @@ func apigeeDynamicDataHeaders(org, env, api string, apiSpec *config.APISpec, fau
 	}
 
 	// Include fault related headers.
-	if fault {
-		headers = append(headers, createHeaderValueOption(headerFaultSource, "ARC", false))
-		headers = append(headers, createHeaderValueOption(headerFaultFlag, "true", false))
-		if apiSpec != nil {
-			headers = append(headers, createHeaderValueOption(headerFaultRevision, apiSpec.RevisionID, false))
+	if adapterFault != nil {
```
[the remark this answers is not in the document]
kansalarchit: Sure, Done.

server/header_context.go
```diff
@@ -104,15 +106,21 @@ func apigeeDynamicDataHeaders(org, env, api string, apiSpec *config.APISpec, fau
...
+		if adapterFault.RpcCode == rpc.INTERNAL {
+			headers = append(headers, createHeaderValueOption(headerFaultSource, "ARC", false))
+			headers = append(headers, createHeaderValueOption(headerFaultFlag, "true", false))
+			if apiSpec != nil {
+				headers = append(headers, createHeaderValueOption(headerFaultRevision, apiSpec.RevisionID, false))
+			}
+
+			if adapterFault.FaultCode == "" {
```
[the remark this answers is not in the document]
kansalarchit: Sure, added a conditional local assignment.

config/env_spec_request.go
```diff
@@ -350,11 +352,11 @@ func (e *EnvironmentSpecRequest) JWTAuthentications() []*JWTAuthentication {
...
 
 // looks up the JWTAuthentication by name and runs verification
-// returns true if found and verified
+// returns nil if found and verified
 // any error can be in e.jwtResults[name]
-func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) bool {
+func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 	if e == nil {
-		return false
+		return fault.CreateAdapterFault("", rpc.UNAUTHENTICATED, 0)
```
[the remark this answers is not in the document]
kansalarchit: Documented that the values can be optional, also added a method that takes just the rpcCode to construct AdapterFault.

server/header_context.go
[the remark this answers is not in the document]
kansalarchit: sure, Done.

config/env_spec_request.go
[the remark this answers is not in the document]
kansalarchit: Sure, documented as well as added a method for CreateAdapterFaultWithRpcCode.

server/authorization.go
```diff
@@ -74,7 +75,7 @@ func (a *AuthorizationServer) Register(s *grpc.Server, handler *Handler) {
 // Check does check
 func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequest) (*authv3.CheckResponse, error) {
 	if !a.handler.Ready() {
-		return a.unavailable(req), nil
+		return a.handleFault(req, nil, nil, "", nil, fault.CreateAdapterFault("", rpc.UNAVAILABLE, 0)), nil
```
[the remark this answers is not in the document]
kansalarchit: 1) I agree, this reduces the readability. This was documented in go/eg4a-runtime-fault-detection and therefore I went ahead and implemented it. Adding call-site param documentation to improve readability.

2) We don't have any use-case, changing the arg to AdapterFault everywhere.

Thanks for the suggestions!

config/env_spec_request.go
```diff
@@ -350,11 +352,11 @@ func (e *EnvironmentSpecRequest) JWTAuthentications() []*JWTAuthentication {
 }
 
 // looks up the JWTAuthentication by name and runs verification
-// returns true if found and verified
+// returns nil if found and verified
 // any error can be in e.jwtResults[name]
-func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) bool {
+func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) *fault.AdapterFault {
```
emschwar, line 357: My intention was for this (and similar return values) to be returned as `error`. Although we do not intend to use the `Error()` function from the `error` interface directly, implementing it would nonetheless allow more conventional Go here, I think.

theganyo: +1 to using as `error`

server/authorization.go
```diff
@@ -495,38 +602,13 @@ func (a *AuthorizationServer) corsPreflightResponse(
...
-	tracker *prometheusRequestMetricTracker, authContext *auth.Context, api string) *authv3.CheckResponse {
-	return a.createConditionalEnvoyDenied(req, envRequest, tracker, authContext, api, rpc.PERMISSION_DENIED)
+	return a.createEnvoyDenied(envRequest.Request, envRequest, tracker, authContext, api, fault.CreateAdapterFault("", rpc.CANCELLED, typev3.StatusCode_NoContent))
 }
 
-func (a *AuthorizationServer) quotaExceeded(req *authv3.CheckRequest, envRequest *config.EnvironmentSpecRequest,
-	tracker *prometheusRequestMetricTracker, authContext *auth.Context, api string) *authv3.CheckResponse {
-	return a.createConditionalEnvoyDenied(req, envRequest, tracker, authContext, api, rpc.RESOURCE_EXHAUSTED)
+func (a *AuthorizationServer) handleFault(req *authv3.CheckRequest, envRequest *config.EnvironmentSpecRequest,
+	tracker *prometheusRequestMetricTracker, api string, authContext *auth.Context, adapterFault *fault.AdapterFault) *authv3.CheckResponse {
```
emschwar, line 609: The `adapterFault` might be better off passed in here as an `error`, and you can test casting it to `fault.AdapterFault` here, or in the `createConditionalEnvoyDenied`. WDYT?

server/authorization.go
```diff
@@ -74,7 +75,13 @@ func (a *AuthorizationServer) Register(s *grpc.Server, handler *Handler) {
 // Check does check
 func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequest) (*authv3.CheckResponse, error) {
 	if !a.handler.Ready() {
-		return a.unavailable(req), nil
```
emschwar, line 77: After reading this, I tend to agree that the prior versions were more readable. I don't think it's required, but if you wanted to bring them back, I don't think that was an essential part of the original design.

config/env_spec_request_test.go
```diff
@@ -365,12 +365,12 @@ func TestIsAuthenticated(t *testing.T) {
 			envoyReq := testutil.NewEnvoyRequest(http.MethodGet, test.path, nil, nil)
 			req := NewEnvironmentSpecRequest(&testAuthMan{}, specExt, envoyReq)
 
-			if req.IsAuthenticated() {
-				t.Fatalf("IsAuthenticated should be false")
+			if req.IsAuthenticated() == nil {
```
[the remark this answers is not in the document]
rockspore: +1

Or it should return both a bool and error?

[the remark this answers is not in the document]
theganyo: Actually, now I'm not sure why we changed this function. We only call it from one place (I believe)... could not we have left it as a boolean and just generate the error at the caller if the result is false?

config/env_spec_request.go
kansalarchit: I had used error initially but changed it later on due to misunderstanding an earlier comment. Changing it back to `error`. Thanks!

config/env_spec_request.go
```diff
@@ -440,9 +446,9 @@ func mustBeInClaim(value, name string, claims map[string]interface{}) error {
 	return fmt.Errorf("%q not in claim %q", value, name)
 }
 
-// IsAuthenticated returns true if AuthenticatationRequirements are met for the request
-// Returns true if AuthenticatationRequirements are empty or disabled.
-func (e *EnvironmentSpecRequest) IsAuthenticated() bool {
+// IsAuthenticated returns nil if AuthenticatationRequirements are met for the request
+// Returns nil if AuthenticatationRequirements are empty or disabled.
+func (e *EnvironmentSpecRequest) IsAuthenticated() *fault.AdapterFault {
```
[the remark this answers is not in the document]
kansalarchit: Done.

config/env_spec_request.go
```diff
@@ -479,33 +485,33 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
-// returns true if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) bool {
+// returns nil if auth is empty or disabled
+func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) *fault.AdapterFault {
```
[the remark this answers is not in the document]
kansalarchit: Done.

config/env_spec_request_test.go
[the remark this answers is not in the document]
kansalarchit: Renamed it to Authenticate() and replaced the usages as suggested by Eric. 
 I think Authenticate() returning an error is more axiomatic way of doing things instead of caller doing the work simply because there is only one caller `so far`.
Thanks!

server/authorization.go
kansalarchit: I had used error initially but changed it later on due to misunderstanding an earlier comment. Changing it back to error. Thanks!

server/authorization.go
kansalarchit: I think with the call-site documentation in place, having a single handleFault looks good too.

theganyo (approved):
Please fix linter error and rebase with master. Other than that, LGMT.

config/env_spec_request.go
```diff
@@ -479,33 +485,33 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
-// returns true if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) bool {
```
rockspore, line 483: This method name also feels very like it should return a boolean, can we rename it to maybe `verifyAuthenticationRequirements`?

fault/fault.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Error() returns the human-readable string representation of the AdapterFault.
+func (f *AdapterFault) Error() string {
+	if f == nil {
+		return "AdapterFault: Error() called on a nil object"
+	}
+	return fmt.Sprintf("FaultCode:%v, RpcCode:%v, StatusCode:%v", f.FaultCode, f.RpcCode.String(), f.StatusCode.String())
+}
+
+// CreateAdapterFaultWithRpcCode() creates and returns a new AdapterFault with the provided input RpcCode.
```
rockspore, line 46: `()` in the comment is not needed. Same for other places.

fault/fault_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+func TestErrorMessage(t *testing.T) {
+	tests := []struct {
+		desc                 string
+		fault                *AdapterFault
+		expectedErrorMessage string
+	}{
+		{
+			desc:                 "test0",
```
rockspore, line 31: The descriptions are better to be something more descriptive about what you are testing in this case :-)

fault/fault_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		{
+			desc:                 "test2",
+			fault:                nil,
+			expectedErrorMessage: "AdapterFault: Error() called on a nil object",
+		},
+	}
+
+	for _, test := range tests {
+		t.Logf("Running test %v", test.desc)
```
rockspore, line 48: This is not needed. And if you are iterating over tests,  the standard pattern in the loop should be:
```
t.Run(test.Desc, func(t *testing.T) {
    // your existing testing logic
})
```

fault/fault_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestCreateAdapterFault(t *testing.T) {
+	tests := []struct {
+		desc               string
+		fault              *AdapterFault
+		expectedFaultCode  string
+		expectedRpcCode    rpc.Code
+		expectedStatusCode typev3.StatusCode
+	}{
+		{
+			desc:               "test0",
```
rockspore, line 65: Same here.

fault/fault_test.go
```diff
@@ -0,0 +1,105 @@
...
+			desc:               "test1",
+			fault:              CreateAdapterFault("", 1, 0),
+			expectedFaultCode:  "",
+			expectedRpcCode:    1,
+			expectedStatusCode: 0,
+		},
+	}
+
+	for _, test := range tests {
+		t.Logf("Running test %v", test.desc)
```
rockspore, line 81: Same here.

server/authorization.go
```diff
@@ -214,47 +259,109 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 	authContext, err := a.handler.authMan.Authenticate(rootContext, apiKey, claims, a.handler.apiKeyClaim)
 	switch err {
 	case auth.ErrNoAuth:
-		return a.unauthenticated(req, envRequest, tracker, api), nil
+		return a.handleFault(
```
rockspore, line 262: line breaking for arguments is not recommended in go. Please convert it to the one-liner even if it's very long.

rockspore: Let me know if I am wrong though ;-)

config/env_spec_request.go
kansalarchit: sure, Done.

fault/fault.go
kansalarchit: Thanks for informing, done.

fault/fault_test.go
kansalarchit: sure, updated.

fault/fault_test.go
kansalarchit: Thanks for pointing out, Done.

fault/fault_test.go
kansalarchit: Done.

fault/fault_test.go
kansalarchit: Done.

server/authorization.go
kansalarchit: Thanks for pointing out. 
From what I can find (refer: https://stackoverflow.com/a/34848928) my use of line-breaks works just fine and it makes code much more readable.

rockspore:
The linting failed so please fix it. Otherwise LGTM. Not approving it right now because once you update the PR the old approval will become ineffective.

kansalarchit:
Fixed linter and resolved all comments, Please take a look. Thanks!

server/authorization.go
```diff
@@ -495,38 +602,13 @@ func (a *AuthorizationServer) corsPreflightResponse(
...
-	tracker *prometheusRequestMetricTracker, authContext *auth.Context, api string) *authv3.CheckResponse {
-	return a.createConditionalEnvoyDenied(req, envRequest, tracker, authContext, api, rpc.PERMISSION_DENIED)
-}
-
-func (a *AuthorizationServer) quotaExceeded(req *authv3.CheckRequest, envRequest *config.EnvironmentSpecRequest,
-	tracker *prometheusRequestMetricTracker, authContext *auth.Context, api string) *authv3.CheckResponse {
-	return a.createConditionalEnvoyDenied(req, envRequest, tracker, authContext, api, rpc.RESOURCE_EXHAUSTED)
+func (a *AuthorizationServer) handleFault(req *authv3.CheckRequest, envRequest *config.EnvironmentSpecRequest,
+	tracker *prometheusRequestMetricTracker, api string, authContext *auth.Context, err error) *authv3.CheckResponse {
+	log.Errorf("Sending fault %v", err.Error())
```
rockspore, line 610: Why err.Error()? err is sufficient for %v

kansalarchit: I assumed that printing the error string might be better. I can change it if you prefer, let me know.

rockspore: IIUC, they are the same. So err.Error() is redundant.

kansalarchit: I see, done.

rockspore (approved):

theganyo (dismissed):

config/env_spec_request_test.go
```diff
@@ -40,10 +43,10 @@ func TestNilReceivers(t *testing.T) {
 	s.GetAPISpec()
 	s.GetOperation()
 	s.GetParamValue(APIOperationParameter{})
-	s.IsAuthenticated()
-	s.verifyJWTAuthentication("")
+	_ = s.Authenticate()
+	_ = s.verifyJWTAuthentication("")
 	s.getAuthenticationRequirement()
-	s.meetsAuthenticatationRequirements(AuthenticationRequirement{})
+	_ = s.verifyAuthenticatationRequirements(AuthenticationRequirement{})
```
emschwar, line 49: nit: this would be a good time to fix this typo (should read: `verifyAuthenticationRequirements`) :joy:

server/authorization.go
emschwar: If the others are fine with is, I am too. Thanks for taking a look!

server/authorization.go
```diff
@@ -133,20 +154,38 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
...
+				envRequest,
+				tracker,
+				api,
+				/* authContext= */ nil,
+				fault.CreateAdapterFaultWithRpcCode(rpc.NOT_FOUND)), nil
 		}
 		log.Debugf("operation: %s", operation.Name)
 
-		if !envRequest.IsAuthenticated() {
+		if authnErr := envRequest.Authenticate(); authnErr != nil {
```
emschwar, line 167: nit: idiomatic Go would use `err` here.

emschwar (dismissed):
A few minor nitpicks, but overall LGTM

server/authorization.go
kansalarchit: Done.

config/env_spec_request_test.go
kansalarchit: haha..thanks for pointing out, Done!

rockspore (dismissed):

config/env_spec_request.go
```diff
@@ -350,11 +352,11 @@ func (e *EnvironmentSpecRequest) JWTAuthentications() []*JWTAuthentication {
 }
 
 // looks up the JWTAuthentication by name and runs verification
-// returns true if found and verified
+// returns nil if found and verified
```
emschwar, line 355: nit: returning nil on success is assumed with Go; maybe rewrite this (and others) as "return error if not found, or not verified"

config/env_spec_request.go
```diff
@@ -440,10 +446,10 @@ func mustBeInClaim(value, name string, claims map[string]interface{}) error {
 	return fmt.Errorf("%q not in claim %q", value, name)
 }
 
-// IsAuthenticated returns true if AuthenticatationRequirements are met for the request
-// Returns true if AuthenticatationRequirements are empty or disabled.
-func (e *EnvironmentSpecRequest) IsAuthenticated() bool {
-	return e.meetsAuthenticatationRequirements(e.getAuthenticationRequirement())
+// Authenticate returns nil if AuthenticatationRequirements are met for the request
+// Returns nil if AuthenticatationRequirements are empty or disabled.
```
emschwar, line 450: Ditto here-- also,  you have two lines that basically say the same thing. Pick one? :)

config/env_spec_request.go
```diff
@@ -479,33 +485,33 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
-// returns true if auth is empty or disabled
-func (e *EnvironmentSpecRequest) meetsAuthenticatationRequirements(auth AuthenticationRequirement) bool {
+// returns nil if auth is empty or disabled
```
emschwar, line 488: Ditto here.

fault/fault.go
```diff
@@ -0,0 +1,62 @@
...
+// Error returns the human-readable string representation of the AdapterFault.
+func (f *AdapterFault) Error() string {
+	if f == nil {
+		return "AdapterFault: Error() called on a nil object"
+	}
+	return fmt.Sprintf("FaultCode:%v, RpcCode:%v, StatusCode:%v", f.FaultCode, f.RpcCode.String(), f.StatusCode.String())
+}
+
+// CreateAdapterFaultWithRpcCode creates and returns a new AdapterFault with the provided input RpcCode.
+func CreateAdapterFaultWithRpcCode(rpcCode rpc.Code) *AdapterFault {
```
emschwar, line 47: Nit: Conventional Go would use `New` vs `Create` here, and below.

emschwar (approved):
A few minor nitpicks here and there, but basically, it looks good to me!

config/env_spec_request.go
kansalarchit: Done

config/env_spec_request.go
kansalarchit: Rephrased.
The two statements are different. First says, nil if met and second says empty/disabled is also considered met.

config/env_spec_request.go
kansalarchit: removed the comment, private function.

fault/fault.go
kansalarchit: Thanks for the info, done.

kansalarchit:
Approvals expired, please take a look again.

server/authorization.go
emschwar: @rockspore is correct here; prefer to not break lines unless it is absolutely required. Our style prefers refactoring to reduce line length, if required, over breaking lines solely for readability.

server/authorization.go
```diff
@@ -214,47 +259,109 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
...
 			return a.authOK(req, tracker, authContext, api, envRequest), nil
 		} else {
-			return a.internalError(req, envRequest, tracker, err), nil
+			return a.handleFault(
+				req,
+				envRequest,
+				tracker,
+				/* api= */ "",
+				/* authContext= */ nil,
+				fault.NewAdapterFaultWithRpcCode(rpc.INTERNAL)), nil
```
emschwar, line 307: Codecov indicates these lines are not covered by a test. Can you add a test, please?

server/authorization.go
```diff
@@ -535,10 +617,16 @@ func (a *AuthorizationServer) quotaExceeded(req *authv3.CheckRequest, envRequest
 func (a *AuthorizationServer) createConditionalEnvoyDenied(
 	req *authv3.CheckRequest, envRequest *config.EnvironmentSpecRequest,
 	tracker *prometheusRequestMetricTracker, authContext *auth.Context,
-	api string, code rpc.Code) *authv3.CheckResponse {
+	api string, err error) *authv3.CheckResponse {
+
+	adapterFault, ok := err.(*fault.AdapterFault)
+	if !ok {
```
emschwar, line 623: This branch doesn't seem to have any tests that cover it. Can you please add one?

emschwar (commented):
Please add a few tests, and I think we can omit breaking lines for readability purposes here. Generally, if we must have such long lines, the solution is to refactor them in order to not be so long, rather than to break the lines up solely for readability purposes.

theganyo (approved):

server/authorization.go
kansalarchit: Done.

server/authorization.go
kansalarchit: This is currently not reachable, its just a safety check we have put in place.

kansalarchit:
As per offline discussion with Eric, Removing the line-breaking call-site documentation and going back to refactored use-case specific methods for `unavailable()` and `internalError()`.
Leaving the rest as it is because it doesn't make any difference for readability.
Also added the unit-test as suggested.

Please take a look. Thanks!

server/header_context.go
```diff
@@ -104,15 +106,19 @@ func apigeeDynamicDataHeaders(org, env, api string, apiSpec *config.APISpec, fau
 	}
 
 	// Include fault related headers.
-	if fault {
+	if adapterFault != nil && adapterFault.RpcCode == rpc.INTERNAL {
```
emschwar, line 109: Hrm, you know you can define a method on `AdapterFault` here that can respond to `nil`? Something like:

```
func (af *AdapterFault) GetRpcCode() rpc.Code {
  if af == nil {
    return rpc.OK
  }

  return af.RpcCode
}
```

(Not saying you have to, but you can simplify these methods a bit if you like)

server/header_context.go
```diff
@@ -104,15 +106,19 @@ func apigeeDynamicDataHeaders(org, env, api string, apiSpec *config.APISpec, fau
...
 		headers = append(headers, createHeaderValueOption(headerFaultSource, "ARC", false))
 		headers = append(headers, createHeaderValueOption(headerFaultFlag, "true", false))
 		if apiSpec != nil {
 			headers = append(headers, createHeaderValueOption(headerFaultRevision, apiSpec.RevisionID, false))
 		}
 
-		// A placeholder fault code value.
-		headers = append(headers, createHeaderValueOption(headerFaultCode, "fault", false))
+		if adapterFault.FaultCode == "" {
```
emschwar, line 116: Like here, you could define an accessor on `AdapterFault` that returns `"fault"` (or "defaultFaultCode", or whatever you like) if there is no `FaultCode` defined, and the value of `FaultCode` otherwise

emschwar (approved):
LGTM. I saw some places you could simplify things a bit, if you feel the need.

server/header_context.go
rockspore: Just wanted to point out it's probably better to do `RpcCode()` since `GetXYZ` is not recommended. I know there are bad examples in this repo as well as the proto-generated pkgs though.

emschwar: Oh gosh, yes! Thanks, I forgot that `GetFoo()` is an antipattern in Go.

kansalarchit: Thanks for the suggestions!
I think method and field with same name cannot co-exist. Skipping this for now.

server/header_context.go
kansalarchit: Thanks for the suggestion, I am not sure if I'd want adapterFault to return `fault` as the default value. The only reason I am keeping this here is to keep the existing logic unchanged.
Skipping this for now.

Issue #371, opened 2022-01-04
Update Fault Codes in x-apigee-fault-code header to contain the MP like fault codes


kansalarchit:
As per https://github.com/apigee/apigee-remote-service-envoy/issues/364, the fault handling was made dynamic.
Logging this issue to populate the actual dynamic fault code values in the x-apigee-fault-code field.

PR #372, opened 2022-01-10
Update Apigee Fault Codes for ARC

kansalarchit:
https://github.com/apigee/apigee-remote-service-envoy/issues/371

Adding x-apigee-fault-code values for key management and JWT based authn and authz as well as other internal errors for ARC.

Notes:
- We cannot get more fine-grained fault-codes for key management yet because of how MP communicates the errors. MP only says [invalid apikey](https://github.com/apigee/apigee-remote-service-cli/blob/master/cmd/provision/proxies/remote-service-gcp/apiproxy/policies/Raise-Fault-Invalid-API-Key.xml). We need to either expand the existing handler or wait for the ACMS apis for getting more details, both of which are beyond the scope of this PR.
- Adding a `default` case to the Key management error `switch-case` as per discussion with @emschwar to ensure new errors in future get handled more gracefully.
- I am not adding / updating the unit tests because looks like we only validate whether the error was raised and not the exact details such as rpc code or fault string for the errors. Let me know if that should change.

### DIFF
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -449,19 +449,18 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
 }
 
 func adapterFaultForJwtErr(err error) *fault.AdapterFault {
-	if err == nil {
+	switch {
+	case err == nil:
 		return fault.NewAdapterFault(fault.JwtUnknownException, rpc.UNAUTHENTICATED, 0)
-	}
-
-	if errors.Is(err, ErrUnsupportedJwkSource) {
+	case errors.Is(err, ErrUnsupportedJwkSource):
 		return fault.NewAdapterFault(fault.JwtInvalidToken, rpc.UNAUTHENTICATED, 0)
-	} else if errors.Is(err, ErrMissingIssuerInClaim) {
+	case errors.Is(err, ErrMissingIssuerInClaim):
 		return fault.NewAdapterFault(fault.JwtIssuerMismatch, rpc.UNAUTHENTICATED, 0)
-	} else if errors.Is(err, ErrMissingAudienceInClaim) {
+	case errors.Is(err, ErrMissingAudienceInClaim):
 		return fault.NewAdapterFault(fault.JwtAudienceMismatch, rpc.UNAUTHENTICATED, 0)
-	} else if errors.Is(err, ErrJwtParsingFailure) {
+	case errors.Is(err, ErrJwtParsingFailure):
 		return fault.NewAdapterFault(fault.JwtInvalidToken, rpc.UNAUTHENTICATED, 0)
-	} else {
+	default:
 		return fault.NewAdapterFault(fault.JwtUnknownException, rpc.UNAUTHENTICATED, 0)
 	}
 }
--- a/config/env_spec_request.go
+++ b/config/env_spec_request.go
@@ -22,6 +22,8 @@ import (
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


server/authorization.go
```diff
@@ -149,7 +149,7 @@ func (a *AuthorizationServer) Check(ctx gocontext.Context, req *authv3.CheckRequ
 			log.Debugf("no authorization requirements")
 			if err := envRequest.PrepareVariables(); err != nil {
 				log.Errorf("failed to populate context variable: %v", err)
```
rockspore, line 151: Let's not log at error level unless there is a fault code. This is related to the ticket I brought up this morning.

kansalarchit: Just to point out, this change has not been done as part of this or earlier PR related to ARC faults.
Could you explain the reason for removing this logging now ?

rockspore: Sorry this is a wrong line to comment. I meant to point out the log.Errorf in `handleFault()` method

kansalarchit: I might not have enough context here but could you share more details on why would we not want to log if there is no specific fault code ? You mentioned that service will be overwhelmed but why do we expect errors to be that many in the first place ?
Also, even that log statement did exist for some of the errors before these changes, since I consolidated the different handleX() methods, it is now there for all of them.

rockspore: It's not EA's error if there is a 40x and we usually only log at debug level for those. It's not uncommon for clients to send bad requests and get 404, while we have access log for metrics purpose, the binary itself does not log those 404s as errors. If you look at Envoy's info log, it shows basically nothing related to the traffic it handles.

Could you point out for me what errors were being logged previously that got consolidated here?

kansalarchit: unavailable() and internalError() were logging. Ref:https://github.com/apigee/apigee-remote-service-envoy/pull/363/files#diff-dc5119ef55fed479882d6318bb13250ddea13884bfdb6aafa3f3359b0e433280L518

So, should we downgrade this to debug log ? or do conditional logging for ARC only ?

rockspore: yeah those two cases are good ones to log errors since it's ARC's problem. The rest should only be debug logs. I believe in those two cases, the fault code will not be empty right? So we just need to conditionally log errors there. The rest can go to debug for troubleshooting purposes.

theganyo: In general, I believe the question is: Is logging XXX useful for support? We should log at >= INFO for cases where that is true.

config/env_spec_request.go
```diff
@@ -423,11 +445,29 @@ func (e *EnvironmentSpecRequest) verifyJWTAuthentication(name string) error {
...
-	return fault.NewAdapterFaultWithRpcCode(rpc.UNAUTHENTICATED)
+	return adapterFaultForJwtErr(err)
+}
+
+func adapterFaultForJwtErr(err error) *fault.AdapterFault {
+	if err == nil {
+		return fault.NewAdapterFault(fault.JwtUnknownException, rpc.UNAUTHENTICATED, 0)
+	}
+
+	if errors.Is(err, ErrUnsupportedJwkSource) {
```
theganyo, line 456: Might this be easier to read as a single switch/case statement?

kansalarchit: thanks for the suggestion, replaced with switch.

server/authorization.go
kansalarchit: downgraded to log.debug level.

theganyo (approved):

PR #377, opened 2022-02-02
Remove x-apigee-* headers before sending to target

### LINKED ISSUE #376: Remove X-Apigee-* headers when forwarding to target

theganyo:
Closes #376

### DIFF
--- a/server/authorization.go
+++ b/server/authorization.go
@@ -295,7 +295,9 @@ func (a *AuthorizationServer) authOK(
 	envRequest *config.EnvironmentSpecRequest) *authv3.CheckResponse {
 
 	checkResponse := a.createEnvoyForwarded(req, tracker, authContext, api, envRequest)
-	checkResponse.GetOkResponse().Headers = append(checkResponse.GetOkResponse().Headers, createHeaderValueOption(headerAuthorized, "true", false))
+	if envRequest == nil {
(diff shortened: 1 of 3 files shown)


server/authorization.go
```diff
@@ -433,17 +435,27 @@ func addRequestHeaderTransforms(req *authv3.CheckRequest, envRequest *config.Env
...
-
 			// header transforms
-			for _, t := range transforms.HeaderTransforms.Remove {
-				t = strings.ToLower(t)
-				for hdr := range req.Attributes.Request.Http.Headers {
-					if util.SimpleGlobMatch(t, hdr) {
-						okResponse.HeadersToRemove = append(okResponse.HeadersToRemove, hdr)
+			for hdr := range req.GetAttributes().GetRequest().GetHttp().GetHeaders() {
+				// strip all x-apigee-* request headers before target
+				if strings.HasPrefix(hdr, "x-apigee-") {
```
rockspore, line 441: We are currently relying on the `x-apigee-message-id` request header in the access log entry to generate the same response header. Have we confirmed that removing this here will not make that header absent from the log entry?

theganyo: The directives to add response headers are based on the incoming request from ingress. These directives to remove headers only affects the outgoing request to the target - and only after all ARC processing is complete.

rockspore (approved):

PR #379, opened 2022-02-03
Update fault codes to better reflect config model.

### LINKED ISSUE #378: Update fault codes to better align with configuration model.
The existing fault codes in fault/fault_codes.go do not reflect the configuration model well. We should update the codes to better match our configuration, so that they'll make more sense to those who consume them.

emschwar:
* Added a new code to better distinguish "api key not found" from
  "api key not present."

Closes #378

### DIFF
--- a/fault/fault_codes.go
+++ b/fault/fault_codes.go
@@ -25,6 +25,6 @@ const (
 	InternalQuotaError            = "consumerAuthorization.quota.InternalError"
 	JwtUnknownException           = "authentication.jwt.InternalError"
 	JwtInvalidToken               = "authentication.jwt.InvalidToken"
-	JwtIssuerMismatch             = "authentication.jwt.JwtIssuerMismatch"
-	JwtAudienceMismatch           = "authentication.jwt.JwtAudienceMismatch"
+	JwtIssuerMismatch             = "authentication.jwt.IssuerMismatch"
+	JwtAudienceMismatch           = "authentication.jwt.AudienceMismatch"
 )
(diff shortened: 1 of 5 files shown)


kansalarchit (commented):
I thought we were going to re-use the existing Apigee MP fault codes, is that not the case anymore ?

fault/fault_codes.go
```diff
@@ -15,15 +15,16 @@
...
-	JwtAudienceMismatch           = "steps.jwt.JwtAudienceMismatch"
+	UnknownAPIProxy               = "environment.ApiProxyNotFound"
+	UnknownKeyManagementException = "consumerAuthorization.InternalError"
+	AuthorizationCodeNotFound     = "consumerAuthorization.RequestMissingKey"
+	InvalidAuthorizationCode      = "consumerAuthorization.UnknownKey"
+	InternalError                 = "apiProxy.InternalError"
+	NoApiProductMatchFound        = "consumerAuthorization.NoApiProductMatchFound"
+	OperationQuotaExceeded        = "consumerAuthorization.quota.LimitExceeded"
+	InternalQuotaError            = "consumerAuthorization.quota.InternalError"
+	JwtUnknownException           = "authentication.jwt.UnknownException"
```
KyleWiese, line 26: authentication.jwt.InternalError

KyleWiese (changes requested):

fault/fault_codes.go
emschwar: Done.

emschwar:
> I thought we were going to re-use the existing Apigee MP fault codes, is that not the case anymore ?

We've decided those codes would be more confusing than helpful at this point.

KyleWiese (dismissed):

theganyo:
Should we be checking that our faults correctly get added to the request? For example, the headers here: https://github.com/apigee/apigee-remote-service-envoy/blob/master/server/authorization_test.go#L343

kansalarchit (approved):

emschwar:
@theganyo I added some tests for the fault code headers; PTAL.

theganyo (approved):

theganyo:
Thanks!

kansalarchit (dismissed):

theganyo (approved):

PR #387, opened 2022-02-18
allow setting dynamic metadata in env_spec

### LINKED ISSUE #386: add support for setting dynamic metadata in env_spec

theganyo:
Closes #386

### DIFF
--- a/config/env_spec.go
+++ b/config/env_spec.go
@@ -253,6 +253,9 @@ type APISpec struct {
 	// Can be overridden at the operation level.
 	ContextVariables []ContextVariable `yaml:"context_variables,omitempty" mapstructure:"context_variables,omitempty"`
 
+	// DynamicMetadata
+	DynamicMetadata map[string]interface{} `yaml:"dynamic_metadata,omitempty" mapstructure:"dynamic_metadata,omitempty"`
+
 	// JWTAuthentication.Name -> *JWTAuthentication
 	jwtAuthentications map[string]*JWTAuthentication `yaml:"-" mapstructure:"-"`
 }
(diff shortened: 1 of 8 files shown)


config/env_spec_request.go
```diff
@@ -524,6 +525,20 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
+func (e *EnvironmentSpecRequest) GetDynamicMetadata() (metadata map[string]interface{}) {
+	if e != nil {
+		op := e.GetOperation()
+		if op != nil && len(op.DynamicMetadata) > 0 {
```
rockspore, line 531: Because this is a map, when absent from the yaml, it may actually nil. Therefore, if users actually specify an empty map in an operation:
```
...
dynamic_metadata: {}
...
```
should we regard that as overriding the API level settings?

config/env_spec.go
```diff
@@ -278,6 +281,9 @@ type APIOperation struct {
 	// Completely overrides the API (proxy) level settings. No merging will happen.
 	ContextVariables []ContextVariable `yaml:"context_variables,omitempty" mapstructure:"context_variables,omitempty"`
 
+	// DynamicMetadata to be given to Envoy. If specified, overrides any DynamicMetadata specified at the API level.
+	DynamicMetadata map[string]interface{} `yaml:"dynamic_metadata,omitempty" mapstructure:"dynamic_metadata,omitempty"`
```
rockspore, line 285: Do we have any use cases that will have something other than strings? If not, should we restrict this map[string]string first?

config/env_spec_request.go
rockspore: This wouldn't handle the case of an empty map at the op level, no?

config/env_spec_request.go
```diff
@@ -524,6 +525,20 @@ func (e *EnvironmentSpecRequest) GetHTTPRequestTransforms() (transforms HTTPRequ
 	return transforms
 }
 
+func (e *EnvironmentSpecRequest) GetDynamicMetadata() (metadata map[string]interface{}) {
```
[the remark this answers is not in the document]
theganyo: done

config/env_spec.go
rockspore: Makes sense. Thanks.

rockspore (approved):

PR #389, opened 2022-02-28
Add support for analytics for gRPC APIs

### LINKED ISSUE #388: Add support for analytics for gRPC APIs.
We need to add support for analytics for gRPC APIs by adding `grpc_status_code` and `grpc_method` fields to the generated json.

MaxSchweizer:
Closes #388

### DIFF
--- a/server/accesslog.go
+++ b/server/accesslog.go
@@ -126,13 +126,21 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
 
 		var api string
 		var authContext *auth.Context
+		var grpcService string
+		var operation string
 
 		extAuthzMetadata := getMetadata(extAuthzFilterNamespace)
 		if extAuthzMetadata != nil {
-			api, authContext = a.handler.decodeAuthMetadata(extAuthzMetadata.GetFields())
+			metadata := a.handler.decodeAuthMetadata(extAuthzMetadata.GetFields())
+			if metadata != nil {
+				api, authContext, grpcService, operation = metadata.Api, metadata.Ac, metadata.GrpcService, metadata.Operation
+			} else {
+				//TODO: What do I do here...
+				log.Debugf("Ut oh")
+			}
 		} else if a.handler.appendMetadataHeaders { // only check headers if knowing it may exist
 			log.Debugf("No dynamic metadata for ext_authz filter, falling back to headers")
-			api, authContext = a.handler.decodeMetadataHeaders(req.GetRequestHeaders())
+			api, authContext = a.handler.decodeMetadataHeaders(req.GetRequestHeaders()) //TODO: Should I change anything here...
 		} else {
 			log.Debugf("No dynamic metadata for ext_authz filter, skipped accesslog: %#v", req)
 			continue
@@ -194,6 +202,8 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
 			ResponseStatusCode:           responseCode,
 			GatewaySource:                a.gatewaySource,
 			ClientIP:                     req.GetForwardedFor(),
+			GrpcStatusCode:               v.GetResponse().GetResponseHeaders()["grpc-status"],
+			GrpcMethod:                   grpcMethod(grpcService, operation),
 			Attributes:                   attributes,
 		}
 
@@ -209,6 +219,14 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
 
 	return nil
 }
+
+func grpcMethod(grpcService string, operation string) string {
+	if grpcService == "" && operation == "" {
+		return ""
+	}
+	return grpcService + "." + operation
+}
+
 
 // returns ms since epoch
 func pbTimestampToApigee(ts *timestamp.Timestamp) int64 {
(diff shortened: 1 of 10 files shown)


server/accesslog.go
```diff
@@ -126,10 +126,17 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
...
+		var grpcService string
+		var operation string
 
 		extAuthzMetadata := getMetadata(extAuthzFilterNamespace)
 		if extAuthzMetadata != nil {
-			api, authContext = a.handler.decodeAuthMetadata(extAuthzMetadata.GetFields())
+			metadata, err := a.handler.decodeAuthMetadata(extAuthzMetadata.GetFields())
+			if err != nil {
+				log.Debugf(err.Error())
+				continue
```
MaxSchweizer, line 137: Having the continue here prevent the `log.Debugf...` on line 149 of the changed version from running.  Im not sure if I should remove the continue or not.

theganyo:
Reminder: Do not merge before https://github.com/apigee/apigee-remote-service-golib/pull/119 is merged and go.mod/go.sum hashes are updated to match.

server/accesslog.go
```diff
@@ -126,10 +126,17 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
...
+			}
+			api, authContext, grpcService, operation = metadata.Api, metadata.Ac, metadata.GrpcService, metadata.Operation
```
emschwar, line 139: Although Go allows parallel assignment like this, it isn't idiomatic. Generally, you'd do them individually, or create some sort of struct type that would contain the results, and build it inline, here. I'd suggest taking a look and see if you really need these as individual fields, or if you can just operate on `metadata` directly for the rest of this func.

server/accesslog.go
```diff
@@ -210,6 +219,14 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
 	return nil
 }
 
+func grpcMethod(grpcService string, operation string) string {
+	if grpcService == "" && operation == "" {
+		return ""
+	}
+	return grpcService + "." + operation
```
emschwar, line 226: You can avoid creating multiple temporary strings here-- consider something like:

```
return fmt.Sprintf("%s.%s", grpcService, operation)
```

There is also `strings.Builder`, but that's probably overkill here, where there's a small number of strings, and they're combined in a fixed way.

server/metadata_context.go
```diff
@@ -36,11 +37,27 @@ const (
 	headerEnvironment    = "x-apigee-environment"
 	headerOrganization   = "x-apigee-organization"
 	headerScope          = "x-apigee-scope"
+	headerGrpcService    = "x-apigee-grpcservice"
+	headerOperation      = "x-apigee-operation"
 )
 
+type AuthMetadata struct {
```
emschwar, line 44: This looks very nice!

server/metadata_context_test.go
```diff
@@ -87,6 +92,43 @@ func TestEncodeMetadataNilCheck(t *testing.T) {
 	}
 }
 
+func TestEncodeMetadatagRPCCheck(t *testing.T) {
```
emschwar, line 95: I don't think you need to downcase the `g` here; although `gRPC` is the proper initialism in writing, I think you're better off treating it as a word, and writing `TestEncodeMetadataGrpcCheck` here.

emschwar (changes requested):

server/accesslog.go
MaxSchweizer: Good point.  I have updated the code per your suggestion.

server/accesslog.go
MaxSchweizer: After looking at how the rest of the code is used, I found that `func (h *Handler) decodeMetadataHeaders(headers map[string]string) (string, *auth.Context)` followed the same return structure that I refactored in `decodeAuthMetadata`.  Having the api and authContext as what is returned causes these values to need to be individual fields.

Similar to `decodeMetadataHeaders`, I have updated this PR so `decodeMetadataHeaders` now returns `(*AuthMetadata, error)`.  This refactor causes both methods to follow the same pattern for decoding, so I was able to refactor out the use of individual fields.

server/accesslog.go
theganyo: Hmm. Maybe we could just consolidate all these err checks and related logging to after this if/else block?

MaxSchweizer: I pushed a changed that consolidates the err checks and related logging to after this if/else block.

server/accesslog.go
```diff
@@ -194,6 +201,8 @@ func (a *AccessLogServer) handleHTTPLogs(msg *als.StreamAccessLogsMessage_HttpLo
 			ResponseStatusCode:           responseCode,
 			GatewaySource:                a.gatewaySource,
 			ClientIP:                     req.GetForwardedFor(),
+			GrpcStatusCode:               v.GetResponse().GetResponseHeaders()["grpc-status"],
```
emschwar, line 197: These failures seem to indicate you need to update the go.mod? Please try it again.

emschwar: Nevermind, I see that this must have been a stale error message. Please ignore.

emschwar (approved):

PR #391, opened 2022-03-01
Go jose

### LINKED ISSUE #390: move JWT processing to go-jose library
Replace [lestratt-go](https://github.com/lestrrat-go/jwx) library with [go-jose](https://github.com/square/go-jose).

theganyo:
Closes #390 
Depends on https://github.com/apigee/apigee-remote-service-golib/pull/121


go.mod
```diff
@@ -35,26 +34,20 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
 
+require gopkg.in/square/go-jose.v2 v2.6.0
```
rockspore, line 37: nit: Why is this not wrapped in the same require above?

server/authmanager.go
```diff
@@ -108,18 +106,30 @@ func (a *JWTAuthManager) stop() {
 func (a *JWTAuthManager) replaceJWT(privateKey *rsa.PrivateKey, kid string, jwtExpiration time.Duration) error {
 	log.Debugf("setting internal JWT")
 
-	token, err := NewToken(jwtExpiration)
+	rsaSigner, err := jose.NewSigner(
```
rockspore, line 109: go doesn't recommend this type of line breaking.

It should instead be
```
rsaSigner, err := jose.NewSigner(jose.SigningKey{
    Algorithm: joes.RS256,
    ...
}, ...)
```

rockspore, line 109: Also, seems like each auth manager is only handling one set of private key and kid, what do you think of a refactoring that reuses this signer?

testutil/jwt.go
```diff
@@ -33,44 +31,33 @@ func GenerateKeyAndJWKs(kid string) (privateKey *rsa.PrivateKey, jwksBuf []byte,
...
+		Keys: []jose.JSONWebKey{jwk},
 	}
 
-	jwksBuf, err = json.MarshalIndent(key, "", "")
+	jwksBuf, err = json.MarshalIndent(jwks, "", "")
 
 	return
 }
 
 func GenerateJWT(privateKey *rsa.PrivateKey, claims map[string]interface{}) (string, error) {
```
rockspore, line 49: nit: Function comment to get rid of one linter error ;)

server/authmanager.go
theganyo: I think it's unnecessary, there is only one AuthManager.

theganyo: done

theganyo: Sure. This is just out of scope of this work, so if we want to track, I think another ticket would be best - but also, this will likely be replaced when we have a formal API instead of the proxy anyway.

rockspore (approved):

emschwar (approved):

PR #399, opened 2023-03-08
Fixed Issue with handling errors like 401, 403, 429 etc and also added new header custom attribute.

mtalreja16:
Fixed Issue with handling errors like 401, 403, 429 etc, the stats were not getting reported back to apigee analytics


theganyo:
We need to break apart the bug fixes from the new feature. I went ahead and created new PRs for the bug fix (https://github.com/apigee/apigee-remote-service-golib/pull/129, https://github.com/apigee/apigee-remote-service-envoy/pull/401) since I wanted to make a couple of simple changes anyway... please take a look. Would you please separate out just your enhancement and make a PR just for that?

PR #402, opened 2023-03-08
Added new header to capture customattributes

mtalreja16:


theganyo (changes requested):

server/metadata_context.go
```diff
@@ -46,15 +47,16 @@ func encodeExtAuthzMetadata(api string, ac *auth.Context, authorized bool) *stru
...
+		headerAccessToken:      stringValueFrom(ac.AccessToken),
+		headerAPI:              stringValueFrom(api),
+		headerAPIProducts:      stringValueFrom(strings.Join(ac.APIProducts, ",")),
+		headerApplication:      stringValueFrom(ac.Application),
+		headerClientID:         stringValueFrom(ac.ClientID),
+		headerDeveloperEmail:   stringValueFrom(ac.DeveloperEmail),
+		headerEnvironment:      stringValueFrom(ac.Environment()),
+		headerOrganization:     stringValueFrom(ac.Organization()),
+		headerScope:            stringValueFrom(strings.Join(ac.Scopes, " ")),
+		headerCustomAttributes: stringValueFrom(ac.CustomAttributes),
```
theganyo, line 59: This header should be only be added if `ac.CustomAttributes` contains a value.

theganyo (changes requested):

server/metadata_context.go
mtalreja16: ahhh make sense...

server/metadata_context.go
```diff
@@ -56,6 +57,11 @@ func encodeExtAuthzMetadata(api string, ac *auth.Context, authorized bool) *stru
 		headerOrganization:   stringValueFrom(ac.Organization()),
 		headerScope:          stringValueFrom(strings.Join(ac.Scopes, " ")),
 	}
+
+	if ac.CustomAttributes != "" {
```
theganyo, line 61: Just one more quick thing: Could you please add a test to verify that this header isn't added when there is no value?

theganyo (approved):

theganyo:
Thank you!

